PR #7, opened 2023-06-22
ON-14982: Use a fixed tag for sfptpd

tcrawley-xilinx:
We want to avoid using `latest` as the only tag for our images.

I am using the commit hash directly here as the most recent tag in [public sfptpd repo](https://github.com/Xilinx-CNS/sfptpd/tags) doesn't have some of the new features we use in the container (specifically `--nodaemon` command line)

## Testing done
tested alongside #3 
sfptpd builds and daemonset runs

### DIFF
--- a/sfptpd/0000-sfptpd-build.yaml
+++ b/sfptpd/0000-sfptpd-build.yaml
@@ -27,17 +27,20 @@ spec:
   strategy:
     dockerStrategy:
       buildArgs:
+      - name: SFPTPD_VERSION
+        value: ab881b3650c642f4dc8142c9b5052da3e6046cdf
   output:
     to:
       kind: ImageStreamTag
-      name: sfptpd:latest
+      name: sfptpd:git-ab881b3
   source:
     dockerfile: |
       FROM registry.access.redhat.com/ubi9/ubi-minimal AS build
       WORKDIR src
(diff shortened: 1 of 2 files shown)


tcrawley-xilinx:
^ Rebased and changed the image name to `git-ab881b3` which should avoid confusion with the image checksum.

sfptpd/0000-sfptpd-build.yaml
```diff
@@ -27,17 +27,20 @@ spec:
...
+      name: sfptpd:git-ab881b3
   source:
     dockerfile: |
       FROM registry.access.redhat.com/ubi9/ubi-minimal AS build
       WORKDIR src
       RUN microdnf -y install make gcc tar bzip2 redhat-rpm-config
       RUN microdnf -y --enablerepo=ubi-9-baseos-source download --source libmnl libcap
-      ADD https://github.com/Xilinx-CNS/sfptpd/archive/refs/heads/master.tar.gz sfptpd.tar.gz
+      ARG SFPTPD_VERSION=master
+      ADD https://github.com/Xilinx-CNS/sfptpd/archive/${SFPTPD_VERSION}.tar.gz sfptpd.tar.gz
```
abower-amd, line 43: If you are labeling with a specific version and have to determine in advance what that version is then it would be better to pull that exact version not a branch thought to have it at that time.

PR #11, opened 2023-06-26
ON-14964: Apply commonLabels to Onload resources

ivatet-amd:
Also, remove the "app" label from a couple of Onload resources.

### DIFF
--- a/onload/base/kustomization.yaml
+++ b/onload/base/kustomization.yaml
@@ -8,3 +8,5 @@ resources:
 - ../userbuild/
 - ../cplane/
 - ../deviceplugin/
+commonLabels:
+  app.kubernetes.io/name: onload
--- a/onload/deviceplugin/0000-buildconfig.yaml
+++ b/onload/deviceplugin/0000-buildconfig.yaml
@@ -3,8 +3,6 @@
 apiVersion: build.openshift.io/v1
 kind: BuildConfig
 metadata:
-  labels:
-    app: onload-device-plugin
   name: onload-device-plugin
 spec:
   nodeSelector:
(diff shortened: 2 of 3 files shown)


abower-amd:
> Not sure how far we want to go with the labels now. We could label the UL build with the container runtime version, e.g. is it el8- or el9-based build of Onload? We could also add a label with the Onload version, but it seems too early, as there are no automated consumers of these like the rolling upgrade story to verify we set the labels correctly.

To my mind the most important version-related metadata that is needed (I'll avoid technical terms I might get wrong; label, version annotation!) are:

- Onload version
- Integration version (when we improve the files in this repo)
- Kernel version (for kernel module builds)

Maybe it is too early for some of this but the fact we are basing off the latest commit of currently unreleased branches for all of the AMD components suggests that this is quite important now - having to delete everything between changes seems to be a lot of friction for development and even less ideal when sharing pre-release versions with lead users.

However, I can see that putting in version information in the wrong way that doesn't actually help with dev/pre-release workflows in any practical sense may be premature, pending better understanding.

pcolledg-xilinx:
If we're removing `app:` from:

* `onload/deviceplugin/0000-buildconfig.yaml`
* `onload/userbuild/buildconfig.yaml`

Why not also remove `app:` from `./onload/deviceplugin/1000-deviceplugin.yaml`? Replace with `name:` as per official doc, or even better `app.kubernetes.io/component:` from Recommended Labels doc.

Same for sfptpd, which also features `app:`.

This variation is presumably due to inconsistencies in official Kubernetes documentation examples.

The `dev` overlay could also include a reference to itself, eg. in `onload/dev/kustomization.yaml`:

```yaml
commonLabels:
  environment: dev
```

ivatet-amd:
Closing in favour of work in [ON-15021](https://jira.xilinx.com/browse/ON-15021).

PR #12, opened 2023-06-27
ON-14920: Reduce example pods capabilities

ivatet-amd:
#### [1/3] ON-14920: Fix SELinux context for Onload files

Set the `container_file_t` type for all files in `/opt/onload` in the compute nodes from the device plugin DS.

The device plugin DS copies Onload files from the UL build image into the node filesystem, ready for use in the user pods when they request the relevant resources with `resources.limits.amd.com/onload: 1`. However, SELinux refuses to grant access to these files to unprivileged pods because they have a wrong `type` in the SELinux `user:role:type:level`. The pods have the `container_file_t` type, but the Onload files have the `var_t` type.

To address the mismatch, the device plugin DS runs `chcon` to set the type in `/opt/onload` recursively.

Note that the `chcon` changes are not permanent and will be reset in the next autorelabel with SELinux. This patch operates on the assumption that autorelabel is a heavyweight operation and occurs only as part of node maintenance, i.e., outside our DS's lifespan. If we discover that this assumption is wrong, we should consider using `semanage fcontext` for permanent changes. It may require a more involved fix, like a new MachineConfig because it implies changes to the node system files, i.e. in `/etc/selinux/`.

#### [2/3] ON-14920: Reduce example pods capabilities

The previous patch that addresses SELinux allows us to reduce the pod capabilities.

#### [3/3] cleanup: Remove Onload files from nodes on device plugin exit

---

Testing notes.

Before applying the patches, the (unprivileged) example pods could not access the files (pod shell):
```console
sh-4.4$ ls -l /opt/onload/usr/lib64/
total 0
-????????? ? ? ? ?            ? libonload.so
-????????? ? ? ? ?            ? libonload_ext.so
-????????? ? ? ? ?            ? libonload_ext.so.1
-????????? ? ? ? ?            ? libonload_ext.so.1.2.0
```

The node audit logs confirm the issue is SELinux-context related (node shell):

```console
# ausearch -m avc --start recent
----
time->Tue Jun 27 14:52:32 2023
type=PROCTITLE msg=audit(1687877552.264:593): proctitle=2F7573722F62696E2F636F72657574696C73002D2D636F72657574696C732D70726F672D73686562616E673D6C73002F7573722F62696E2F6C73002D6C002F6F70742F6F6E6C6F61642F7573722F6C696236342F
type=SYSCALL msg=audit(1687877552.264:593): arch=c000003e syscall=257 success=no exit=-13 a0=ffffff9c a1=7eff34bc00f0 a2=80000 a3=0 items=0 ppid=459748 pid=459963 auid=4294967295 uid=1000 gid=0 euid=1000 suid=1000 fsuid=1000 egid=0 sgid=0 fsgid=0 tty=pts0 ses=4294967295 comm="ls" exe="/usr/bin/coreutils" subj=system_u:system_r:container_t:s0:c161,c625 key=(null)
type=AVC msg=audit(1687877552.264:593): avc:  denied  { read } for  pid=459963 comm="ls" name="libonload.so" dev="vda4" ino=15066390 scontext=system_u:system_r:container_t:s0:c161,c625 tcontext=system_u:object_r:var_t:s0 tclass=file permissive=0
```

After applying the patches (pod shell):
```
sh-4.4$ ls -l /opt/onload/usr/lib64/
total 12460
-rwxr-xr-x. 1 root root 12672720 Jun 27 16:57 libonload.so
-rwxr-xr-x. 1 root root    27416 Jun 27 16:57 libonload_ext.so
-rwxr-xr-x. 1 root root    27416 Jun 27 16:57 libonload_ext.so.1
-rwxr-xr-x. 1 root root    27416 Jun 27 16:57 libonload_ext.so.1.2.0
```

Also, since the pods are no longer privileged, OCP does not issue warnings (`oc` user shell):
```console
$ oc apply -f examples/cns-sfnettest.yaml
imagestream.image.openshift.io/onload-sfnettest unchanged
buildconfig.build.openshift.io/onload-sfnettest configured
pod/onload-sfnettest-server created
pod/onload-sfnettest-client created
```

### DIFF
--- a/examples/cns-sfnettest.yaml
+++ b/examples/cns-sfnettest.yaml
@@ -51,8 +51,9 @@ metadata:
 spec:
   restartPolicy: Always
   securityContext:
-    allowPrivilegeEscalation: false
-    runAsUser: 0
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
   containers:
   - name: onload-sfnettest-server
     image: image-registry.openshift-image-registry.svc:5000/default/onload-sfnettest:latest
@@ -63,7 +64,10 @@ spec:
       limits:
         amd.com/onload: 1
     securityContext:
-      privileged: true
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+          - ALL
   nodeName: compute-0
 ---
 apiVersion: v1
@@ -76,8 +80,9 @@ metadata:
 spec:
   restartPolicy: Never
   securityContext:
-    allowPrivilegeEscalation: false
-    runAsUser: 0
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
   containers:
   - name: onload-sfnettest-client
     image: image-registry.openshift-image-registry.svc:5000/default/onload-sfnettest:latest
@@ -88,5 +93,8 @@ spec:
       limits:
         amd.com/onload: 1
     securityContext:
-      privileged: true
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+          - ALL
   nodeName: compute-1
--- a/onload/deviceplugin/1000-deviceplugin.yaml
+++ b/onload/deviceplugin/1000-deviceplugin.yaml
@@ -65,7 +65,13 @@ spec:
         lifecycle:
           postStart:
             exec:
-              command: ["cp", "-TRv", "/opt/onload", "/host/onload"]
+              command:
+              - /bin/sh
+              - -c
+              - |
+                set -e;
+                cp -TRv /opt/onload /host/onload;
+                chcon --verbose --type container_file_t --recursive /host/onload/;
       volumes:
         - hostPath:
             path: /var/lib/kubelet/device-plugins
--- a/onload/deviceplugin/1000-deviceplugin.yaml
+++ b/onload/deviceplugin/1000-deviceplugin.yaml
@@ -72,6 +72,14 @@ spec:
                 set -e;
                 cp -TRv /opt/onload /host/onload;
                 chcon --verbose --type container_file_t --recursive /host/onload/;
+          preStop:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - |
+                set -e;
+                rm -r /opt/onload /host/onload;
       volumes:
         - hostPath:
             path: /var/lib/kubelet/device-plugins


examples/cns-sfnettest.yaml
```diff
@@ -51,8 +51,10 @@ metadata:
 spec:
   restartPolicy: Always
   securityContext:
-    allowPrivilegeEscalation: false
-    runAsUser: 0
+    runAsUser: 1000
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
```
abower-amd, line 57: I'm surprised some of this isn't the default? Same for later snippets.

ivatet-amd: OC demands (some of) these security fields in the manifests. I ran a small experiment to verify if all of these were needed with `oc apply` with and without `--dry-run=server` and discovered a couple we can remove! So, thanks for raising the concern. 👍

- Without `spec.securityContext.runAsNonRoot: true`:
```
Warning: would violate PodSecurity "restricted:latest": runAsNonRoot != true (pod or container "onload-sfnettest-client" must set securityContext.runAsNonRoot=true)
```

- Without `spec.securityContext.seccompProfile.type: RuntimeDefault`:
```
Warning: would violate PodSecurity "restricted:latest": seccompProfile (pod or container "onload-sfnettest-client" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

- Without `spec.containers.[].securityContext.capabilities.drop: [ALL]`:
```
Warning: would violate PodSecurity "restricted:latest": unrestricted capabilities (container "onload-sfnettest-client" must set securityContext.capabilities.drop=["ALL"])
```

- Without `spec.containers.[].securityContext.allowPrivilegeEscalation: false`:
```
Warning: would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "onload-sfnettest-client" must set securityContext.allowPrivilegeEscalation=false)
```

- Without `spec.securityContext.runAsUser: 1000`: no complaints!
- Without `spec.containers.[].securityContext.privileged: false`: no complaints!

Additionally, it matches with the template pod suggested by OCP UI in the "Create pod" view:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: example
  labels:
    app: httpd
  namespace: default
spec:
  securityContext:
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
  containers:
    - name: httpd
      image: 'image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest'
      ports:
        - containerPort: 8080
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
```

I'm removing these unneeded fields from the manifests in the upcoming force push.

abower-amd (approved):

onload/deviceplugin/1000-deviceplugin.yaml
```diff
@@ -65,7 +65,13 @@ spec:
...
           postStart:
             exec:
-              command: ["cp", "-TRv", "/opt/onload", "/host/onload"]
+              command:
+              - /bin/sh
+              - -c
+              - |
+                set -e;
+                cp -TRv /opt/onload /host/onload;
+                chcon --verbose --type container_file_t --recursive /host/onload/;
```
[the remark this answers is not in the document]
ivatet-amd: Thanks for the comprehensive comment.

> How about `cp --context=container_file_t ...`

Thanks, I only used `--preserve=context` in the past. However, it doesn't do what we want from within a device plugin pod:
```console
sh-4.4# cp -v --context='system_u:object_r:container_file_t:s0' /host/onload/foo.txt /host/onload/bar.txt
cp: warning: ignoring --context; it requires an SELinux-enabled kernel
'/host/onload/foo.txt' -> '/host/onload/bar.txt'
```

It didn't apply the SELinux labels, and to understand why I ran it with `strace`:
```console
# strace cp -v --context='system_u:object_r:container_file_t:s0' /host/onload/foo.txt /host/onload/bar.txt
...
access("/etc/selinux/config", F_OK)     = -1 ENOENT (No such file or directory)
prctl(PR_SET_NAME, "cp")                = 0
prctl(PR_SET_KEEPCAPS, 140726852443438) = -1 EINVAL (Invalid argument)
geteuid()                               = 0
write(2, "cp: ", 4cp: )                     = 4
write(2, "warning: ignoring --context; it "..., 66warning: ignoring --context; it requires an SELinux-enabled kernel) = 66
```

It looks like our base image is not SELinux-friendly. But I'm not keen to choose another image only for that purpose!

> A slightly more persistent version would be

Yes, and I tried to explain that in the commit/PR message: it implies changes to the host file system, which we probably should do with MC rather than from within a pod, and it seems too heavy-weight for this occasion.

> a simple initContainer for the file transfer

IMO, this is also heavy-weight because we would have to "mount" another image merely for a couple of shell commands.

> - an ./entrypoint.sh for the device plugin container, or
> - performing the file transfer within device plugin Go code itself.

Interesting. Either of these two options is better than `postStart` alone, at least because they would keep the copying logs. (The lifecycle hooks are dull, exposing the logs only on failures in the form of an event. 🤦‍♂️ ) However, this made me realise that we probably want to use the `preStop` to remove the Onload files, and it is somewhat more reliable than the entry point or Golang because it's handled with K8s itself.

> OCP v4.12 has first-class [definitions](https://docs.openshift.com/container-platform/4.12/security/security_profiles_operator/spo-selinux.html) in the form of `SelinuxProfile`.

Do you have insights about that kind, i.e. how to harness it for our use case (copying files to the host FS)?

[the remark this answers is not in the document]
ivatet-amd: TL;DR: I will add the `preStop` lifecycle hook to remove host files, but please comment if you disagree.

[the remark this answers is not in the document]
pcolledg-xilinx: "it requires an SELinux-enabled kernel" is a message from [coreutils cp](https://github.com/coreutils/coreutils/blob/master/src/cp.c#L1170), conditional on the value set by `is_selinux_enabled ()`, a function provided by `libselinux` package. Perhaps that package is missing from the container and it has defaulted to assuming disabled. That package [is in ubi repos](https://catalog.redhat.com/software/containers/ubi8/ubi/5ff3f50a831939b08d1b832a?container-tabs=packages&tag=8.7-8&push_date=1680612950000).

For reference, `libselinux-devel` package contains: `/usr/include/selinux/selinux.h` & `/usr/share/man/man3/is_selinux_enabled.3.gz`.

---

>> a simple initContainer for the file transfer
>
> IMO, this is also heavy-weight because we would have to "mount" another image

The onload-user image is already consumed on the node, and this would be a per-device-plugin container not per-user-pod, so very low cost. Hopefully avoidable anyway.

---

Details of applying `SelinuxProfile` are in the link, under "Applying SELinux profiles to a pod", however we cannot explore this while v4.10 is considered supported. Ideally we should define a custom SELinux Context (eg. `onload_container_file_t`) so the user pod is only exposed to our files.

examples/cns-sfnettest.yaml
pcolledg-xilinx: The extra security overrides are required due to the presence of:
```yaml
metadata:
  namespace: default
```
Removing or otherwise changing the namespace to almost anything else (eg. `onload-runtime`) removes the security violation warning.

onload/deviceplugin/1000-deviceplugin.yaml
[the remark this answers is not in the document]
ivatet-amd: Thanks again.

I did a bit of more research on the topic. Apparently, `libselinux` is already installed with the base:
```console
sh-4.4# rpm -q libselinux
libselinux-2.9-8.el8.x86_64
```

In the `strace` report above, `cp` was complaining about missing `/etc/selinux/config`, so I checked where it comes from:
```console
$ yum whatprovides '/etc/selinux/config' | grep --before-context=3 Filename | head -4
selinux-policy-3.14.1-61.el8.noarch : SELinux policy configuration
Repo        : lab-rhel8.6-x86_64-baseos
Matched from:
Filename    : /etc/selinux/config
```

Then, installed `selinux-policy` in a pod:
```
sh-4.4# microdnf install selinux-policy
```

Retried `cp` still "fails" (exit code is 0):
```console
sh-4.4# cp -v --context='system_u:object_r:container_file_t:s0' /host/onload/foo.txt /host/onload/bar.txt
cp: warning: ignoring --context; it requires an SELinux-enabled kernel
'/host/onload/foo.txt' -> '/host/onload/bar.txt'
```

However, now the `strace` report is slightly different:
```console
sh-4.4# strace cp -v --context='system_u:object_r:container_file_t:s0' /host/onload/foo.txt /host/onload/bar.txt
access("/etc/selinux/config", F_OK)     = 0
prctl(PR_SET_NAME, "cp")                = 0
prctl(PR_SET_KEEPCAPS, 140721982192946) = -1 EINVAL (Invalid argument)
geteuid()                               = 0
write(2, "cp: ", 4cp: )                     = 4
write(2, "warning: ignoring --context; it "..., 66warning: ignoring --context; it requires an SELinux-enabled kernel) = 66
write(2, "\n", 1
)                       = 1
```
I traced a little further with `gdb` and discovered that `is_selinux_enabled()` is checking `has_selinux_config`, which is True, and `selinux_mnt`, which is False. The `selinux_mnt` is set in `init_selinuxmnt()` which is checking that:
- `/sys/fs/selinux` (`SELINUXMNT`) is present in a pod but has a wrong magic value. The `f_type` returned with `statfs` is `0x62656572` (`SYSFS_MAGIC`) while expected to be `0xf97cff8c` (`SELINUX_MAGIC`). https://man7.org/linux/man-pages/man2/statfs.2.html
- `/selinux` (`OLDSELINUXMNT`) is not present.
- `/proc/mounts` (legacy way of discovering SELinux) does not contain a line with SELinux.

I tried to set `shareProcessNamespace`, `hostIPC` and `hostPID` out of curiosity if it helps to peek into the node's procfs, but it doesn't achieve that goal. (I don't want to escape the namespace merely to make the pod SELinux-friendly even if it worked.)

Another thought is that, presumably, the userland SELinux library must be compatible with the kernel's SELinux, and not any base image would be a good fit for the purpose, limiting our choice for the device plugin, which is meant to be no more than a Golang runtime.

Overall, I think `chcon` is the way to go for this exercise.

Re on `initContainer`: I understand it is a low overhead, but I don't see why it's better in the light of `preStop`?

Re on `SelinuxProfile`: ACK. Overall, it looks interesting as it makes SELinux more kube-friendly.

examples/cns-sfnettest.yaml
ivatet-amd: That is a good find, thank you. (Also, for checking in a test cluster with `onload-sfnettest-client-pcolledg-bb9b4b4b6-fhbdd`.)

Let me move them into `onload-examples` and remove the remaining security fields in the upcoming git-force-push.

ivatet-amd: Actually, let me keep that as-is. As discussed with @pcolledg-xilinx offline, the new "onload-examples" won't address the admission problem because new namespaces are created "restricted", and we would need to relax that constraint if we want to remove the security-related fields. Filed ON-15013 to drill into the security aspect more widely.

pcolledg-xilinx: We need to document OCP v4.10 exceptions (ie. remove `seccompProfile:`)

ivatet-amd: OCP v4.10 seems to define `seccompProfile` along with `RuntimeDefault` in https://docs.openshift.com/container-platform/4.10/rest_api/workloads_apis/pod-v1.html.

PR #17, opened 2023-06-29
ON-15008: Allow the use of onload profiles

tcrawley-xilinx:
This can allow a user to run an onloaded application with a profile.
The standard way of using a profile isn't compatible with the `LD_PRELOAD` method of onloading an application so the profile has to convert manually to a file specifying the environment variables (eg `examples/profiles/latency/latency.conf`).

Two example profiles are provided `latency` and `throughput-base` which are based on the corresponding profiles in the onload directory.

The profiles are applied using kustomization to patch the pod definitions.

### Misc changes
* The example pods now live in `examples/sfnettest/` and are deployed using kustomize
* I have removed the example output from `examples/sfnettest/README.md` since the command line invocation and numbers are now out of date, also I think it looks a bit ugly. If people think it is worthwhile then I will be happy to add an equivalent with new numbers.

## Testing done
All of:
`oc create -k examples/sfnettest`
`oc create -k examples/profiles/latency/`
`oc create -k examples/profiles/throughput-base`
Are tested and work in the cluster.


README.md
```diff
@@ -96,7 +96,7 @@ The server pod is already running the accelerated `sfnt-pingpong` instance.
 
 Run the client:
 ```
-sh-4.4# LD_PRELOAD=/opt/onload/usr/lib64/libonload.so ./sfnt-pingpong udp 198.19.0.1
+sh-4.4# ./sfnt-pingpong udp 198.19.0.1
 ```
 
 Try running with and without `onload` to compare the reported performance.
```
abower-amd, line 102: To run without onload you either need to remove the resource specifier in the yaml or prefix command with `LD_PRELOAD= `, which would be simpler for a quick test, I think?

README.md
```diff
@@ -96,7 +96,7 @@ The server pod is already running the accelerated `sfnt-pingpong` instance.
 
 Run the client:
 ```
-sh-4.4# LD_PRELOAD=/opt/onload/usr/lib64/libonload.so ./sfnt-pingpong udp 198.19.0.1
+sh-4.4# ./sfnt-pingpong udp 198.19.0.1
```
abower-amd, line 99: Can we make sure this main top-level example uses the latency profile to get good out-of-the-box performance?

abower-amd (approved):

ivatet-amd:
I like when the PR comes as a series of patches because it adds the time dimension. (And I like Gerrit for encouraging this!)

However, when we do series, _each_ patch must bring us to the correct state of things, e.g. the code must be compilable. So unless you want to squash the PR, I suggest revisiting how the PR organises the patches:

> Use kustomize to deploy example sfnettest pods

We probably need a `cleanup:` or `ON-xyz` prefix here.

> Cleanup: Update oc invocations in READMEs

I suggest squashing this one into the previous because otherwise, the README contain lies about the moved cns-sfnettest.yaml. One exception is the `LD_PRELOAD` line which deserves its cleanup patch.

> ON-15008: Allow the use of onload profiles

I think this patch deserves more than just a subject line! Please explain that we can't consume the OPF files as-is from `/opt/onload/` without modifying the user/accelerated pod, which we want to avoid. However, the upcoming automation with the operator will address it (since we can pre-process the OPF files in Golang and feed them to the pod). Now, we copy-and-paste the files for demo purposes.

> Cleanup: Tidy up the sfnettest README.md

Why are we removing the sample output numbers? Consider replacing it with better latency numbers, i.e. when the profile is applied.

> Cleanup: fix example pods location in READMEs

This should be squashed to the patch that moves the files across. Otherwise, the README contains lies!

README.md
tcrawley-xilinx: I meant to remove that line as well.
Prefixing `LD_PRELOAD=` would only stop the client from being accelerated, it would not affect the server.
I think the neatest way to stop acceleration would be to remove the request for the `amd.com/onload` resource.

ivatet-amd:
This commit:

> ON-15008: Allow the use of onload profiles

... screams for https://cbea.ms/git-commit/#why-not-how 😉

pcolledg-xilinx:
Is a ConfigMap supplying env vars preferable to [supplying a file](https://kubernetes.io/docs/tutorials/configuration/configure-redis-using-configmap/)? Presumably the end result is identical.

tcrawley-xilinx:
> Is a ConfigMap supplying env vars preferable to [supplying a file](https://kubernetes.io/docs/tutorials/configuration/configure-redis-using-configmap/) (.opf)? Presumably the end result is identical.

Currently we can't do anything with the existing onload profile `.opf` files, because they need the `onload` script to process them.
One of the goals for running onload in the cluster with these resources was to have minimal (ideally no) changes to the user application's pod.
Using `LD_PRELOAD` to accelerated an app and handling the profiles by using a ConfigMap to set environment variables requires to fewest changes to the user application pod.

tcrawley-xilinx:
@ivatet-amd 
> > Cleanup: Tidy up the sfnettest README.md
> 
> Why are we removing the sample output numbers? Consider replacing it with better latency numbers, i.e. when the profile is applied.

I addressed this in the main description (under the "misc changes" section).
Basically I think it looks quite ugly / noisy and doesn't add much valuable information. I don't feel particularly strongly about it, so if you think it is worth it I'll add a new section with the good latency numbers.

tcrawley-xilinx:
^ rebased onto new master and consolidated into fewer patches.
Also added a commit description for the reasoning behind the design choices for the profiles

ivatet-amd:
> I think it looks quite ugly / noisy

Why?

> and doesn't add much valuable information.

They show a sample output. They are somewhat verbose, and this is the reason why they reside in an individual README file.

> I don't feel particularly strongly about it, so if you think it is worth it I'll add a new section with the good latency numbers.

`Cleanup: Tidy up the sfnettest README.md` seems to suit well for the updated numbers?

Also, thank you for reviewing the PR patch series, it looks good. 👍

If you don't mind nit-picking:
- `LD_PRELOAD` fixes could be moved from `ON-15015 Use kustomize to deploy example sfnettest pods` to the last patch `Cleanup: Tidy up the sfnettest README.md` because these fixes are "tidy up".
- Suggesting a `:` after here: `ON-15015: Use kustomize`.

tcrawley-xilinx:
^ Updated commit name

ivatet-amd (approved):

PR #23, opened 2023-07-04
ON-14911: Get container ID through controller runtime API

tcrawley-xilinx:
For crictl to start the cplane it needs to get the onload cplane container's ID. The current implementation gets the container ID by checking the cgroup in the container. Apart from being hacky, this solution only worked for cgroup v1.

This change adds a new golang program which communicates with the kuberenetes controller-runtime to get the details of the currently running pod and to update the cplane server parameters. This will work on both cgroup v1 and v2 systems.

## testing done
Builds and runs in container. Tested building from github using a personal fork


onload/cplane/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+	containerID string
+)
+
+func main() {
+
+	podName = os.Getenv("POD_NAME")
+	podNamespace = os.Getenv("POD_NAMESPACE")
+
+	cfg, err := config.GetConfig()
```
pcolledg-xilinx, line 42: Official [example](https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go#L41) uses `rest.InClusterConfig()` from `k8s.io/client-go/rest` library. Is there a reason the controller library is required here instead?

onload/cplane/main.go
```diff
@@ -0,0 +1,133 @@
...
+			time.Sleep(timeoutTime)
+			continue
+		}
+
+		if containerStatus.State.Running == nil {
+			glog.Warningf("Container is not running. Will retry in %v",
+				timeoutTime)
+			time.Sleep(timeoutTime)
+			continue
+		}
```
pcolledg-xilinx, line 93: Go supports nested functions, so this could be DRY by just passing the state string to a err handler `if ... warnAndWait('ready')`

onload/cplane/main.go
```diff
@@ -0,0 +1,133 @@
...
+	err = os.WriteFile(
+		cplaneServerParamsFilePath,
+		[]byte(cplaneServerParams),
+		0644,
+	)
+	if err != nil {
+		glog.Fatalf("Couldn't write cplane_server_params")
+	}
+
+	// We want this program to run indefinetly
```
pcolledg-xilinx, line 131:
```suggestion
	// We want this program to run indefinitely
```

pcolledg-xilinx (commented):
Generally looks good.

I think there is an argument for creating an autonomous open source utility which _just_ outputs the `containerID`, and then utilise it in a specific cplane param writing script, but that can be reconsidered later.

onload/cplane/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+		containerID = containerStatus.ContainerID
+
+		break
+	}
+
+	// The ContainerID reported has a "cri-o://" prefix before the actual ID.
+	// We need to remove this so that crictl is happy.
+	// If the prefix is not found, then we should be good to continue, but log
+	// the event anyway.
```
pcolledg-xilinx, line 103: Under what circumstances would the string be populated but not contain the cri-o schema prefix? I can only think of the case where a different CRI were in use, and this application only supports `critclFilePath` type CRI, therefore is this instead a fatal 'unsupported' exception?

onload/cplane/main.go
pcolledg-xilinx: Under what circumstances would the `containerID` change once found?

(Also, next line `select {}` is redundant?)

tcrawley-xilinx: I don't expect `containerID` to change, the problem is that if the program finishes then kubernetes will stop the DaemonSet and mark it as "Completed" which causes it to start the loop again.

We need a way to keep the pod alive indefinitely (or at least until the pod is removed). I could add something like `&& sleep infinitely` to the end of the command, but that would be ugly IMO.

`select {}` is used so that the program will block forever without eating cpu cycles.

pcolledg-xilinx:
```suggestion
	// Block here indefinitely
```
Ah, I see it is a common hack, thanks. Wasn't how I assumed it'd behave. This would be clearer to me.

onload/cplane/main.go
tcrawley-xilinx: My previous understanding was that the kubernetes controller library was necessary for getting the controller ID, but obviously that is not the case.
I have just tested it using the `client-go` library and it works, so in a second I'll update the PR to use as I think it would be more appropriate.

onload/cplane/main.go
tcrawley-xilinx: I think you're right, I'll update the message and make it fatal

tcrawley-xilinx:
Thoughts on adding logging? currently nothing is output on success

pcolledg-xilinx (approved):
Well there could be a debug line "written '{params}' to {file}" but exiting cleanly should suffice; catting the file isn't hard. We should ensure the file is included in ON-15024.

onload/cplane/main.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		glog.Fatalf("Failed to create clientset (%v)", err)
+	}
+	pod := &corev1.Pod{}
+
+	// The information about the pod may be out of date when it is received,
+	// so we enter a "while" loop to keep checking about the pod and container
+	// until they report as running.
+	for {
+		pod, err = clientSet.CoreV1().Pods(podNamespace).Get(
```
ivatet-amd, line 63: Shall we move the "get container ID(s?) by pod namespace and name" part into a separate function?

onload/cplane/main.go
```diff
@@ -0,0 +1,131 @@
...
+	// We need to remove this so that crictl is happy.
+	// If this is not present then we are using an unknown container runtime
+	// interface, in that case abort.
+	containerID, found := strings.CutPrefix(containerID, crioPrefix)
+	if !found {
+		glog.Fatalf("Could not find %q as a prefix to the containerID %v",
+			crioPrefix, containerID)
+	}
+
+	err = os.WriteFile(
```
ivatet-amd, line 108: I think both `os.WriteFile()` calls each could be a separate function like `setCplaneServerPath()` and `setCplaneServerParams()` or even in OOP-style, a new struct `cplaneServer` with methods `setPath()` and `setParams()`?

ivatet-amd (commented):
I'm up for unit tests before merging... The inline comments may help with unit tests, e.g. split some aspects into new functions and then mock them to test something else.

PR #24, opened 2023-07-05
ON-14945: ON-15027: Separate build and deploy steps to support using pre-built images in air-gapped clusters.

tcrawley-xilinx:
A series of commits that separates the build and deploy steps of sfc, onload, sfptpd and sfnettest. This does add a lot of `kustomization.yaml` files, but it should mean that all of the existing kustomize commands work, but now you have more control about what builds you want to take place (if any).

## Testing done
Tested deploying everything in a cluster and passing accelerated traffic both when build in cluster and when using pre-built images.


README.md
```diff
@@ -162,4 +177,142 @@ $ oc get image | grep onload
 ```
 Remove any outstanding manually with `oc delete image`. (Not providing any automated invocation to prevent removal of the false-positive images.)
 
+## Deploying artifacts into an airgapped cluster
+
+### Images needed
```
pcolledg-xilinx, line 182:
```suggestion
### Images required
```

README.md
```diff
@@ -162,4 +177,142 @@ $ oc get image | grep onload
...
+default-route-openshift-image-registry.apps.test.kube.test/onload-clusterlocal/onload-module         git27b3826-4.18.0-372.49.1.el8_6.x86_64
+default-route-openshift-image-registry.apps.test.kube.test/onload-clusterlocal/onload-device-plugin  latest
+default-route-openshift-image-registry.apps.test.kube.test/onload-clusterlocal/onload-user           git27b3826
+default-route-openshift-image-registry.apps.test.kube.test/openshift-kmm/sfc-module                  git27b3826-4.18.0-372.49.1.el8_6.x86_64
+```
+
+### Getting the images
+
+The images can either be built using podman directly (untested) or using a
+non-airgapped cluster.
```
pcolledg-xilinx, line 199: Restricted networks which have local Git and Container Registry hosting can still use HTTP-based build methods.

```suggestion
The images can either be built using podman directly (currently unsupported) or on a
cluster with access to a git web host, either github.com or a locally hosted clone of <named repos>.
```

README.md
```diff
@@ -162,4 +177,142 @@ $ oc get image | grep onload
...
+
+
+#### Pulling images from a cluster
```
pcolledg-xilinx, line 202: OpenShift sysadmins will be highly proficient in image handling so will not require an introduction to podman. This section would be better served by a ready-to-go script covering all the necessary images.

README.md
```diff
@@ -162,4 +177,142 @@ $ oc get image | grep onload
...
+```
+sudo podman login -u kubeadmin -p $(oc whoami -t) ${REGHOST} --tls-verify=false
+```
+
+#### A Note on `--tls-verify=false`
+
+Without doing any extra steps podman will complain about the certificates used
+by the image registry. For development purposes `--tls-verify=false` can be
+used to bypass this issue. A proper solution would be to extract the
+certificates and pass these to podman.
```
pcolledg-xilinx, line 253:
```suggestion
used to bypass this issue. Production environments shall follow internal certificate handling procedures to secure podman access.
```

README.md
```diff
@@ -162,4 +177,142 @@ $ oc get image | grep onload
...
+```
+
+### Deploying sfc driver
+
+#### KMM deployment
+First remove the in-tree driver from the worker nodes
+```
+# rmmod sfc
+```
+This should allow the kmm to install the kernel module without issue.
```
pcolledg-xilinx, line 280: If this command is required it should be added to `sfc/mco/99-sfc-machineconfig.bu` rather than left as an exercise for the user.

onload/dev/kustomization.yaml
```diff
@@ -6,22 +6,3 @@ namePrefix: dev-
 resources:
 - ../base
 
-patches:
```
pcolledg-xilinx, line 9: What adds `--debug` instead?

pcolledg-xilinx (changes requested):
ImageStreams can go in the build.yamls as they're [otherwise unused](https://xilinx-cns.slack.com/archives/C04J5E8KD5W/p1688033917627779?thread_ts=1687944575.000659&cid=C04J5E8KD5W) and, like BuildConfig, are OpenShift-specific.

README.md
tcrawley-xilinx: Removing `sfc` and `sfc_driverlink` is already handled in [`sfc-replace-kernel-module.sh#L9`](https://github.com/Xilinx-CNS/kubernetes-onload/blob/e556d4d5186cca334ccf56424263f658bd6e6036/sfc/mco/sfc-replace-kernel-module.sh#L9).

This is only necessary since we are using KMM v1.0 (which is the only one I can find on the operator hub). If we were using KMM v1.1 there is the `inTreeModuleToRemove` field which would handle the removal for us.

onload/dev/kustomization.yaml
tcrawley-xilinx: This is just something I forgot to restore from previous testing, I'll add in back in.

README.md
tcrawley-xilinx: I think I'm going to leave this as is for the time being since this document is useful to members of our team who aren't proficient with image handling.

README.md
pcolledg-xilinx: Good point, this is useful information to have in that section of the documentation. 

~~~suggestion
#### KMM Module deployment

If KMM v1.0 is used, the in-tree driver must be removed from the worker nodes
before deployment. This is a one-time action which is automated by the
MachineConfig component upon reboot. In KMM v1.1 onwards, add 
`inTreeModuleToRemove: sfc` to the `sfc` Module CR.

```sh
rmmod sfc
```
~~~

tcrawley-xilinx: Your suggestions looks good, I've added it as a separate commit

pcolledg-xilinx (approved):

README.md
```diff
@@ -162,4 +172,136 @@ $ oc get image | grep onload
 ```
 Remove any outstanding manually with `oc delete image`. (Not providing any automated invocation to prevent removal of the false-positive images.)
 
+## Deploying artifacts into an airgapped cluster
+
+### Images required
+
+This is a list of the images that are currently needed as reported by podman.
+```
```
ivatet-amd, line 180:
```suggestion
```console
```
nit: I recently learnt that \`\`\`console is a better way for a multiline snippets like that (with one command and the rest is output)!

ivatet-amd:
May I suggest considering less generic names for the files named "patch.yaml"?

README.md
```diff
@@ -162,4 +172,136 @@ $ oc get image | grep onload
...
+Without doing any extra steps podman will complain about the certificates used
+by the image registry. For development purposes `--tls-verify=false` can be
+used to bypass this issue. Production environments shall follow internal
+certificate handling procedures to secure podman access.
+
+### Pushing the images to the cluster's image-registry
+
+For each image you want to push run the command:
+```
+$sudo podman push REGISTRY/OPENSHIFT_NAMESPACE/IMAGE_NAME:TAG --tls-verify=false
```
ivatet-amd, line 255:
```suggestion
$ sudo podman push REGISTRY/OPENSHIFT_NAMESPACE/IMAGE_NAME:TAG --tls-verify=false
```

README.md
```diff
@@ -162,4 +172,136 @@ $ oc get image | grep onload
 ```
 Remove any outstanding manually with `oc delete image`. (Not providing any automated invocation to prevent removal of the false-positive images.)
 
+## Deploying artifacts into an airgapped cluster
+
+### Images required
+
+This is a list of the images that are currently needed as reported by podman.
```
ivatet-amd, line 179:
```suggestion
This is a list of the images that are currently needed as reported by podman:
```

I know it's really nit-picking but I can't ignore typography inconsistency! We tend to use : (colon mark) for referring to the code snippets. If you disagree, then choose any approach and stick to it here and elsewhere like "Podman login". 👍

ivatet-amd:
I think we should retire the existing "onload/base" and "onload/dev", otherwise, we are duplicating YAMLs.

The suggested directory tree is:
```
onload/
  build/
    base/
    debug/ (or whatever we call build profiles)
    release/
  deploy/
    base/
    stage/
    prod/
  resources/
    cplane/
    deviceplugin/
    imagestream/
    kmm/
    userbuild/
  clusterlocal/
    stage-debug/ (Kustomization files in these directories should refer to the above "build" and "deploy")
    prod-release/
```
What do you reckon?

README.md
ivatet-amd: How about moving the `podman` sections into a separate README file? The top-level one becomes overwhelming. I bet we should keep it short and friendly!

If you think the main consumers are going to be our team members (who are highly skilled! 😜), how about moving these bits into the internal knowledge base like Confluence?

ivatet-amd:
Another thought about "examples/profiles"...

Profiles are the deployment aspect of the application, i.e. we don't rebuild (and don't want to?) the sfnt-pingpong binary depending on the profile. Thus, why not move the profile files like "examples/profiles/throughput-base/deploy" in "examples/sfnettest/deploy/throughput"?

Also, not sure why we call it "throughput-base" (i.e. "-base")?

README.md
pcolledg-xilinx: This can be part of ON-14978.

tcrawley-xilinx:
Merged. Further naming / structure conversations can continue, but I don't think they prevent this from going in.

PR #25, opened 2023-07-10
ON-14933: Update onload device plugin go dependencies

tcrawley-xilinx:
### [1/2] ON-14933: Update onload device plugin go dependencies

Dependencies were updated using `go get -u`.

`go list -m -u` still shows out of data packages, but these are from
indirect dependencies.

Required a couple of changes to the code:
* `grpc.WithInsecure()` is now deprecated, so use preferred version
instead.
* `DevicePluginServer` interface requires implementing
`GetPreferredAllocation`, so a stub version was added for compliance.

### [2/2] ON-14903: Implement GetDevicePluginOptions for onload device plugin

The onload device plugin does not implement the optional functions
PreStartContainer and GetPreferredAllocation. To prevent the kubelet
requesting something that we can't handle, use DevicePluginOptions to
say that these are unsupported.

## Testing done
Builds and runs in cluster, with accelerated traffic passing between example pods
`staticcheck` and `golangci-lint` report no issue with the code.


pcolledg-xilinx (approved):
Please update versions in `onload/deviceplugin/ContainerFileNotice`.

Side notes on versions:

* Kubelet version corresponds to client-go version and both are tracking latest Kubernetes stable version but [neither](https://github.com/kubernetes/client-go#compatibility-client-go---kubernetes-clusters) have a detailed compatibility matrix.
* The golang tag in `onload/deviceplugin/Dockerfile` is already [latest stable](https://quay.io/repository/projectquay/golang?tab=tags).

PR #27, opened 2023-07-13
ON-15022: Add onload_mkimages script to build necessary images outside of a cluster using podman

tcrawley-xilinx:
### [1/2] ON-15022: Add onload_mkimages script
onload_mkimages.sh takes as input a filepath to a tarball release of
onload as well as some information about the cluster to build for
(OpenShift version, node kernel version) and produces a set of images
from the onload release that should work in the described cluster.

A new set of Dockerfile is also included in the scripts/ directory which
are used by onload_mkimages because the existing Dockerfiles are built
to pull from github, not work with local files.

### [2/2] ON-14945: Pull pre-built images from example external registry

Update the images registry mentioned in the deploy kustomization files
for onload and sfc to use `www.example.com` as the source for the images.

## Testing done
Used the `onload_mkimages.sh` script to build images on a DUT (outside of
cluster) and then pushed to a local image registry, then updated the
kustomization files to pull the images from the registry.
Deploy everything in the cluster and ran onload with accelerated traffic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,10 @@ MachineConfig component upon reboot. In KMM v1.1 onwards, add
 rmmod sfc
 ```
 
+_The kustomization patch file [sfc/deploy/patch-sfc-module-no-build.yaml](sfc/deploy/patch-sfc-module-no-build.yaml)
+will attempt to pull the images from the fictional image registry at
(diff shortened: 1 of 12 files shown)


pcolledg-xilinx:
Let's discuss.

PR #31, opened 2023-07-20
SCL-1739: Add missing ContainerFileNotice

ivatet-amd:


ivatet-amd:
I copied that file from "onload/deviceplugin", but it looks like all *.go files should belong to "onload/build/deviceplugin" to prevent cross-dependencies?

ivatet-amd:
Closing for https://github.com/Xilinx-CNS/kubernetes-onload/pull/32.

PR #32, opened 2023-07-20
ON-15027: Move deviceplugin sources to onload/build/deviceplugin

ivatet-amd:
Address cross-dependency between onload/deviceplugin and onload/build/deviceplugin.

---

Because there is a chicken and egg situation, tested with this patch:

```diff
diff --git a/base/onload-sources.conf b/base/onload-sources.conf
index 73d9e85..f4cd14e 100644
--- a/base/onload-sources.conf
+++ b/base/onload-sources.conf
@@ -1,2 +1,2 @@
 ONLOAD_LOCATION=https://github.com/Xilinx-CNS/onload/archive/refs/tags/v8.1.0.tar.gz
-KUBERNETES_ONLOAD_LOCATION=https://github.com/Xilinx-CNS/kubernetes-onload/archive/refs/heads/master.tar.gz
+KUBERNETES_ONLOAD_LOCATION=https://github.com/Xilinx-CNS/kubernetes-onload/archive/5e80695eb7ade337d7ec13dff742b429a0bd6b3c.tar.gz
```


tcrawley-xilinx:
Changes look good, but I think we will want to hold off from merging until #25 is in.

ivatet-amd:
Rebased after the conflict auto-resolve. Re-testing.

base/onload-sources.conf
```diff
@@ -1,2 +1,2 @@
 ONLOAD_LOCATION=https://github.com/Xilinx-CNS/onload/archive/refs/tags/v8.1.0.tar.gz
-KUBERNETES_ONLOAD_LOCATION=https://github.com/Xilinx-CNS/kubernetes-onload/archive/refs/tags/v3.0.0-pre1.tar.gz
+KUBERNETES_ONLOAD_LOCATION=https://github.com/Xilinx-CNS/kubernetes-onload/archive/refs/heads/master.tar.gz
```
ivatet-amd, line 2: @tcrawley-xilinx @pcolledg-xilinx I'm updating the location because otherwise, the patch would break the device plugin build. We should re-tag it again to pinpoint the device plugin revision.

We have alternatives:
1. We could pick a fixed arbitrary hash (without a tag).
2. We could bump this reference to a non-existing tag, commit, push and then create a new (missing, referred to) tag.
3. Keep using the branch.

Ideally, this would be nice to have minimal automation to address this self-reference. I believe with the real "operator" this problem will go away.

However, in the absence of automation, I prefer the first option because it's simple. (We could also tag the device plugin separately, but it seems excessive.)

The second option seems too fragile from the workflow point of view, though it looks neat from the historical point of view.

The third option seems too fragile, as well as the second option, because this could suddenly break a developer's setup if someone else has pushed the device plugin changes.

I understand this PR brings little value (addresses cross-dependency), but it serves as a good illustration of a problem anyone would come across while working with the device plugin.

Opinions?

tcrawley-xilinx: I think option 2 is the best, and use `master` for now

tcrawley-xilinx (approved):

PR #38, opened 2023-07-25
ON-14945: Make base image locations customisable

tcrawley-xilinx:
This change adds a set of new variables to onload-sources.conf which allows users to customise the location of base images needed to build the set of onload images.

Default values are set to use internet registries.

## Testing done
Everything builds in-cluster (using default values) and accelerated traffic passes.

### DIFF
--- a/onload/kmm/Dockerfile
+++ b/onload/kmm/Dockerfile
@@ -1,8 +1,10 @@
 # SPDX-License-Identifier: MIT
 # SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
 ARG DTK_AUTO
+ARG DTK_IMAGE
+ARG UBI8_IMAGE
 
-FROM ${DTK_AUTO} as builder
+FROM ${DTK_IMAGE:-${DTK_AUTO}} as builder
 ARG ONLOAD_BUILD_PARAMS
 ARG ONLOAD_LOCATION
 ARG KERNEL_VERSION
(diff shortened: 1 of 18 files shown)


onload/kmm/Dockerfile
```diff
@@ -1,8 +1,10 @@
 # SPDX-License-Identifier: MIT
 # SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
 ARG DTK_AUTO
+ARG DTK_IMAGE
+ARG UBI8_IMAGE
 
-FROM ${DTK_AUTO} as builder
+FROM ${DTK_IMAGE:-${DTK_AUTO}} as builder
```
ivatet-amd, line 7: AFAIR, this `FROM` statement might make docker "forget" `ARG UBI8_IMAGE` used in the next `FROM`. Why is it not the case here? Also, noticed the same in diagnostics.

ivatet-amd (commented):
This is quite a neat use of Kustomize! I have one inline question about the ordering of `ARG` and `FROM` in Dockerfile, please take a look.

onload/kmm/Dockerfile
tcrawley-xilinx: My understanding (which could be wrong) is that the `ARG` parameters which happen before the first `FROM` act more like global variables.
https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

ivatet-amd: Thanks, this part of the docs seems to be written specifically for such occasions!

ivatet-amd (approved):

tcrawley-xilinx:
Closing this PR as I don't think this is the right approach.
Rather having a custom solution to update the image locations it would probably be better to utilise existing mechanisms in the cluster.

This [redhat doc](https://docs.openshift.com/container-platform/4.12/openshift_images/image-configuration.html) has more information on ways to get the cluster to pull the image from the right location.

PR #45, opened 2023-08-11
ON-15141: Define Onload CRD

tcrawley-xilinx:
An initial definition of the Onload CRD to be used by the operator. I don't think this PR needs to be all-encompassing, but should allow for progress to be made on the reconciliation loop.

## Open questions
1. How "flat"/"nested" should the definition be?
2. What should go in the Status section?
3. Where should sfptpd go?

## Testing done
Operator image built and pushed into cluster's registry. Then deployed operator with no issues.
Adding and deleting the example CR in `config/samples` works as expected.
Log output
```
2023-08-11T11:03:18Z	INFO	Adding new Onload kind	{"controller": "onload", "controllerGroup": "onload.amd.com", "controllerKind": "Onload", "Onload": {"name":"onload-sample","namespace":"onload-operator-system"}, "namespace": "onload-operator-system", "name": "onload-sample", "reconcileID": "ceb08dab-6b2d-42c1-a1be-8cdf3c1c6c3a"}
2023-08-11T11:04:00Z	INFO	Onload resource not found. Ignoring since object must be deleted	{"controller": "onload", "controllerGroup": "onload.amd.com", "controllerKind": "Onload", "Onload": {"name":"onload-sample","namespace":"onload-operator-system"}, "namespace": "onload-operator-system", "name": "onload-sample", "reconcileID": "804f74c3-daf9-429b-90f1-b3407cc49c71"}
```


api/v1alpha1/onload_types.go
```diff
@@ -4,19 +4,49 @@
...
+	UserImage string `json:"userImage"`
+
+	// DevicePluginImage
+	DevicePluginImage string `json:"devicePluginImage"`
+
+	// +optional
+	// Optional version string to associate with this onload CR.
+	Version string `json:"version,omitempty"`
+
+	// +optioanl
```
ivatet-amd, line 36:
```suggestion
	// +optional
```

ivatet-amd (commented):
Thomas and I had an offline discussion, Thomas is going to update the PR.

api/v1alpha1/onload_types.go
```diff
@@ -4,25 +4,76 @@
...
+	// including the sfc driver.
+	KernelImage string `json:"kernelImage"`
+
+	// UserImage
+	// An image that contains the built userland objects, used by the cplane
+	// and deviceplugin DaemonSets.
+	UserImage string `json:"userImage"`
+
+	// DevicePluginImage
+	DevicePluginImage string `json:"devicePluginImage"`
```
ivatet-amd, line 34: Do we want to move it to the (new) `DevicePluginSpec`?

api/v1alpha1/onload_types.go
```diff
@@ -4,25 +4,76 @@
...
+
+	// +optional
+	// Optional version string to associate with this onload CR.
+	Version string `json:"version,omitempty"`
```
ivatet-amd, line 38: Do we want to move it to the (existing) `Spec`?

api/v1alpha1/onload_types.go
```diff
@@ -4,25 +4,76 @@
...
+}
+
 // OnloadSpec defines the desired state of Onload
 type OnloadSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// KernelImage
+	// An image that contains the out-of-tree kernel modules used by onload,
+	// including the sfc driver.
+	KernelImage string `json:"kernelImage"`
```
ivatet-amd, line 26: Do we want multiple kernel versions like `regexp` in https://kmm.sigs.k8s.io/documentation/deploy_kmod/ to (1) support the node kernel upgrade more seamlessly and (2) allow the selector to select nodes with different kernel versions?

api/v1alpha1/onload_types.go
```diff
@@ -4,25 +4,76 @@
...
+	// +optional
+	// Build contains build options.
+	// If non-nil the operator will attempt to build the required images using
+	// what is defined in the build section, and push the images to the
+	// specified locations.
+	Build *OnloadBuildSpec `json:"build,omitempty"`
+
+	// +optional
+	// Specific instructions for building the device plugin image.
+	DevicePluginBuild *DevicePluginBuildSpec `json:"devicePluginBuild,omitempty"`
```
ivatet-amd, line 54: Do we want to move it to the (new) `DevicePluginSpec` along with `DevicePluginImage`?

pcolledg-xilinx:
Doesn't the use of single `*Image` attributes discard all the flexibility and heuristics provided by KMM, such as kernel version regex selecting?

Previously I've suggested that for user-manageable attribs, we take inspiration from KMM Hub's user-supplied [`moduleSpec:`](https://kmm.sigs.k8s.io/documentation/hub_spoke/#the-managedclustermodule-crd). Maybe start with [ModuleLoaderSpec](https://github.com/rh-ecosystem-edge/kernel-module-management/blob/main/api/v1beta1/module_types.go#L226). Obviously we can iterate, this PR does not need to fully replicate Hub immediately, but using Hub as a skeleton will provide space for complexities likely required soon.

I agree with merging definitions of one-to-one dependent components where possible.

For statuses & metrics, we can start with minimum and iterate -- observability is later on in ON-15116.

As per ON-15116, I bumped `sfptpd` to phase2 milestone.

api/v1alpha1/onload_types.go
tcrawley-xilinx: This was intended to be the version of onload used by the operator (eg 8.1.0, enterprise 8 etc.)
We could additional information in the top-level `Spec` struct, but this is onload specific.

pcolledg-xilinx: I believe `Version` should be required everywhere for clusters.

Confer context around my proposal:
> Users must manually change the SemVer version to manage an upgrade. The `:latest` tag may still be useful for non-cluster solutions.
>   -- Production Distribution - Draft

ivatet-amd:
Thank you, the CRD looks good to me! Let's see what @pcolledg-xilinx says.

ivatet-amd (approved):

api/v1alpha1/onload_types.go
```diff
@@ -4,25 +4,118 @@
...
+}
+
+type KernelMapping struct {
+	// Regexp is a regular expression that is used to match against the kernel
+	// versions of the nodes in the cluster
+	Regexp string `json:"regexp"`
+
+	// KernelImage is the image that contains the out-of-tree kernel modules
+	// used by onload, including the sfc driver.
+	KernelImage string `json:"kernelImage"`
```
pcolledg-xilinx, line 24: `KernelImage` could be misconstrued as a full kernel not just kernel _module_. I think we should always include the word 'module' or 'driver'. KMM calls this DriverContainer or just ContainerImage, but I'm not partial provided we're consistent.

api/v1alpha1/onload_types.go
```diff
@@ -4,25 +4,118 @@
...
 type OnloadSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
 
-	// Foo is an example field of Onload. Edit onload_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// KernelMappings is a list of pairs of kernel versions and container
+	// images. This allows for flexibility when there are heterogenous kernel
+	// versions on the nodes in the cluster.
+	KernelMappings []KernelMapping `json:"kernelMappings"`
```
pcolledg-xilinx: KMM (inc. KMM Hub, and indeed K8s in general) always defines container config in a sub `container` struct to avoid conflating the CRI abstraction of Pod/DaemonSet config with container engine Container config. We could push this granularity to a later milestone.

pcolledg-xilinx (commented):
Good skeleton

tcrawley-xilinx:
Merged.
The intention of this was to be a starting point that would unblock work on the reconciliation loop and allow for further iterations, and I think I have achieved that.

PR #47, opened 2023-08-16
ON-15121: Create control plane daemon set

ivatet-amd:
A draft until these work items are done:

- [x] Rebase on top of https://github.com/Xilinx-CNS/kubernetes-onload/pull/46.
- [x] Reuse the `ServiceAccountName` field added in both PRs.
- [ ] ~Delete DS manually before deleting Module to facilitate the ordering.~
- [x] Add test coverage: test if the reconciler calls `Create()` only if `Get()` returns `IsNotFound()`.

---

The operator reconciliation spawns the control plane daemon set that registers itself with the kernel module set shortly afterwards:

```console
$ oc apply -k config/samples
$ dmesg
[  +0.000023] [onload] Copyright (c) 2002-2023 Advanced Micro Devices, Inc.
[  +0.003796] [onload] oo_nic_add: ifindex=3046 oo_index=0
[Aug21 09:03] onload_cp_server[2418478]: Onload Control Plane server   started: id 0, pid 2418478
[  +0.003175] onload_cp_server[2418478]: Accelerating sf0: RX 1 TX 1
```

Testing:
```console
$ make test
KUBEBUILDER_ASSETS="/home/iteterev/lab/kubernetes-onload/bin/k8s/1.26.0-linux-amd64" go test ./... -coverprofile cover.out
ok      github.com/Xilinx-CNS/kubernetes-onload/controllers     7.311s  coverage: 74.6% of statements
```

### DIFF
--- a/controllers/onload_controller.go
+++ b/controllers/onload_controller.go
@@ -12,15 +12,27 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	kmm "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 
 	onloadv1alpha1 "github.com/Xilinx-CNS/kubernetes-onload/api/v1alpha1"
+	"k8s.io/utils/ptr"
 )
+
+const controlPlaneScript = `
+set -e;
+echo /usr/bin/crictl | tee /sys/module/onload/parameters/cplane_server_path;
+declare -r container_id=$(awk -F'[-./]' '/crio-/{print $(NF - 1); exit}' /proc/self/cgroup);
+echo exec ${container_id} /opt/onload/sbin/onload_cp_server -K | tee /sys/module/onload/parameters/cplane_server_params;
+sleep infinity;
+`
 
 // OnloadReconciler reconciles a Onload object
 type OnloadReconciler struct {
@@ -32,6 +44,7 @@ type OnloadReconciler struct {
 //+kubebuilder:rbac:groups=onload.amd.com,resources=onloads/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=onload.amd.com,resources=onloads/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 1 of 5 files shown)


ivatet-amd:
Early testing goes well: the operator controller can create a CP DS that updates the Onload sysfs as expected. The deletion of the CP DS happens automatically with K8s (cascading deletion) when the user deletes the Onload CR because the Onload CR is the owner of the CP DS.

internal/utils/utils.go
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+
+package utils
+
+func NewBoolPointer(v bool) *bool {
+	return &v
+}
```
tcrawley-xilinx, line 8: You could use this existing k8s utils package.
https://pkg.go.dev/k8s.io/utils/ptr
Something like `ptr.To[bool](True)`

ivatet-amd: Yep, it's a good idea, thanks.

controllers/onload_controller.go
```diff
@@ -182,6 +203,72 @@ func createModule(
...
+		ctx,
+		types.NamespacedName{
+			Name:      dsName,
+			Namespace: onload.Namespace,
+		},
+		ds,
+	)
+	if err == nil || !apierrors.IsNotFound(err) {
+		return err
+	}
```
tcrawley-xilinx, line 223: Am I missing something or is there a logic bug here? returning `err` when `err == nil` is just returning `nil` which presumably we don't want to do in an error path

ivatet-amd: This is "if DS already exists (err is nil) OR there is a genuine error (not IsNotFound) then return either nil (the former case) or the error (the latter case)". That's what I think we do with the module kind handing too, though I see it's expressed differently.

I see how it's convoluted and will split these checks into two with a bit of logging!

Still, this is a temporary approach that must be improved with something like https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch to patch the existing DS if it doesn't match the desired spec.

tcrawley-xilinx: Ok thanks, I hadn't considered that the if statement was a combination of two genuinely different cases.
Some logging / comments would be nice

controllers/onload_controller.go
```diff
@@ -56,15 +85,190 @@ func (r *OnloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+
+	container := v1.Container{
+		Name:            onload.Name + cpdsPodSuffix,
+		Image:           onload.Spec.Onload.UserImage,
+		ImagePullPolicy: v1.PullAlways,
```
tcrawley-xilinx, line 228:
```suggestion
		ImagePullPolicy: onload.Spec.Onload.ImagePullPolicy
```

controllers/onload_controller.go
ivatet-amd: Done, please take a look at the new code. 👍

controllers/onload_controller.go
ivatet-amd: Done.

tcrawley-xilinx (approved):

pcolledg-xilinx (approved):
Assuming rebase passes & this being temporary. (Would prefer controlPlaneScript in a separate file if this must be kept in longer term.)

ivatet-amd:
> Assuming rebase passes & this being temporary.

That's right: I want to check in the control plane DS as-is from the PoC branch to enable the Onloaded application as early as possible, i.e. after this PR and https://github.com/Xilinx-CNS/kubernetes-onload/pull/49 have merged. I will upgrade it to one of the API approaches from [ON-14911](https://jira.xilinx.com/browse/ON-14911) immediately after.

> Would prefer controlPlaneScript in a separate file if this must be kept in longer term.

Yes, I initially wanted to add the "internal" directory with (at least) control-plane constants, e.g. suffixes we append to the DS ("-onload-cplane-ds") and pod name ("-onload-cplane"), along with the control plane script. This would: (1) isolate the operand-specific bits and pieces from the reconciliation loop, and (2) provide some nice namespacing, e.g. `control_plane.Script` or `control_plane.DaemonSetSuffix`.

In the end, I was convinced it's probably not worth creating a new module merely for that + the script will be removed soon™.

Please let me know if that's still OK with you?

PR #49, opened 2023-08-21
ON-15119: Add device plugin

tcrawley-xilinx:
~~Still in progress, there is some code copied from #47 which will have to removed once that is merged~~ Done
### Patch 1/3
Adds support for the onload device plugin to the controller using existing image from `v3.0.0-pre`.
### Patch 2/3
Adds the source code for the device plugin from `v3.0.0-pre` with some minor modifications so it can handle being in a different directory structure.
Also added new commands to the makefile to build and push the device plugin image
### Patch 3/3
Updates the onload controller to use the images produced by patch 2
### Patch 4/3
Separates the device plugin into `pkg/deviceplugin` and `cmd/deviceplugin` to better align with the [golang recommended project layout](https://github.com/golang-standards/project-layout).
 - `pkg/deviceplugin` still contains the main logic for the device plugin, but now in a package called `deviceplugin`
 - `cmd/deviceplugin` contains the `main` function for the device plugin and acts as an entrypoint into the package.
#### Automated test
```console
$ make test
KUBEBUILDER_ASSETS="/home/tcrawley/Dev/kubernetes-onload/bin/k8s/1.26.0-linux-amd64" go test ./... -coverprofile cover.out
?       github.com/Xilinx-CNS/kubernetes-onload/pkg/deviceplugin        [no test files]
ok      github.com/Xilinx-CNS/kubernetes-onload/controllers     8.221s  coverage: 77.3% of statements
```

#### Manual testing
 * Built both device plugin image and controller image and pushed to dev registry.
 * Deployed operator using `make deploy`
 * Added Onload CR using `oc create -k config/samples` & verified all Modules/DaemonSets were created
 * Tested device plugin
   * Created an example pod requesting an `onload` resource -- succeeds with the device plugin active and fails otherwise
   * Deployed sfnettest pods from `v3.0.0-pre` and verified accelerated traffic was passing
  * Deleted Onload CR (after deleting user pods) to verify everything was removed


deviceplugin.Dockerfile
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+FROM golang:1.20.4 AS builder
+
+COPY go.mod /app/go.mod
+COPY go.sum /app/go.sum
+COPY pkg/deviceplugin /app/
```
ivatet-amd, line 7: A plain Dockerfile is so much easier than using BuildConfig.

Makefile
```diff
@@ -133,6 +135,14 @@ docker-build: test ## Build docker image with the manager.
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
 
+.PHONY: device-plugin-build
+device-plugin-build: test ## Build docker image with the manager.
+	docker build -t ${DEVICE_IMG} -f deviceplugin.Dockerfile --network="host"
```
ivatet-amd, line 140: Not sure if @pcolledg-xilinx has suggestions about what to call "deviceplugin.Dockerfile" and where to put it?

ivatet-amd (approved):

Makefile
tcrawley-xilinx: According the golang recommended project layout there should be a `/build/package` directory with all the packaging related stuff there.
https://github.com/golang-standards/project-layout/blob/d0cc0aa49671e3d65add3755b7a14f2df6b12374/build/README.md?plain=1#L5
How closely we follow the recommendation is a different question

deviceplugin.Dockerfile
```diff
@@ -2,12 +2,16 @@
...
 COPY go.sum /app/go.sum
-COPY pkg/deviceplugin /app/
+RUN go mod download
 
-WORKDIR /app
-RUN go build -o /app/onload-plugin
+COPY pkg/deviceplugin /app/pkg/deviceplugin
+COPY cmd/deviceplugin /app/cmd/deviceplugin
+
+RUN go build -o /app/onload-plugin ./cmd/deviceplugin
```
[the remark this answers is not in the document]
tcrawley-xilinx: The existing Dockerfile for the controller calls `go build` directly ([link](https://github.com/Xilinx-CNS/kubernetes-onload/blob/fb64119340d506dd02fd880933c4e06f7d2768fc/Dockerfile#L27)) which I think is preferable as the `build` target in the makefile has at least `manifests generate fmt vet` as prerequisites, which will increase build time / complexity.

I acknowledge that `build` and `device-plugin-build` will be different make targets, but I like the idea of mirroring the functionality between them.

tcrawley-xilinx:
@pcolledg-xilinx 
I apparently can't reply to [this comment](https://github.com/Xilinx-CNS/kubernetes-onload/pull/49#discussion_r1302817720) in-line, so I am responding here.

>  I anticipated this kubernetes-onload repo would become two top-level directories

That was the initial idea, but after looking at implementing this change moving from the current directory structure it became unappealing.

> Looks like the primary extraneous dep the device-plugin will inherit is kmm lib, and rpc for the operator

I think this should be fine as the unused stuff in `go.mod` and `go.sum` won't be compiled/linked, so the only real downside is that the `go mod download` steps in the dockerfiles will take slightly longer. We could add a `go mod tidy` step to the dockerfile, but that would require re-ordering the dockerfile which would make the builds perform worse with the docker cache.

PR #52, opened 2023-08-30
[WIP] ON-15173: Label nodes on which onload runs

tcrawley-xilinx:
This change does a few things:
 * add a step to the reconciliation loop to label nodes which match the provided selector.
 * Set the selector of the created Modules and Daemonsets to run on nodes that label.
 * Gets the controller to watch for changes of nodes, and run the reconciliation loop for the onload CR in case they can run on the node after the changes.
 * On deletion of an onload CR remove the labels from the nodes

Not sure if this is the right approach going forwards, but I wanted it to be visible.

## Testing done
Manual testing in the cluster. 

#### Test 1
 * I created an onload CR that only runs on a single node.
 * Then created an additional onload CR which could run on all nodes
 * Verified that only a single onload CR was running on each node.
 * Removed the first onload CR
 * Verified the labels for the first CR were removed.
 * Verified that onload was now running on the second node.

#### Test 2
 * Created an onload CR with a selector that matches a label on no nodes.
 * Verified no pods were created
 * Add a label to a node
 * Verified that pods were created for this node
 * Add a label to the second node
 * verified pods were created on the second node

#### Known issue
If a node no longer matches the onload CR's selector it won't remove the onload label automatically, this means that pods will continue running until the label is removed manually.

### DIFF
--- a/controllers/onload_controller.go
+++ b/controllers/onload_controller.go
@@ -6,6 +6,7 @@ package controllers
 import (
 	"context"
 	"errors"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -18,7 +19,9 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	kmm "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 
@@ -45,6 +48,7 @@ type OnloadReconciler struct {
 //+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=onload.amd.com,resources=onloads/finalizers,verbs=update
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -70,6 +74,11 @@ func (r *OnloadReconciler) Reconcile(
 			// In this way, we will stop the reconciliation
 			log.Info("Onload resource not found." +
 				" Ignoring since object must be deleted")
+			err := r.deleteLabels(ctx, req.NamespacedName)
+			if err != nil {
+				log.Error(err, "Failed to clean labels after deletion")
+				return ctrl.Result{}, err
+			}
 			return ctrl.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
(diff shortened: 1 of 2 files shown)


controllers/onload_controller.go
```diff
@@ -108,6 +122,84 @@ func (r *OnloadReconciler) Reconcile(
...
+
+	for _, node := range nodes.Items {
+		nodeCopy := node.DeepCopy()
+		labels := node.GetObjectMeta().GetLabels()
+		for k := range labels {
+			if strings.HasPrefix(k, "onload.amd.com") {
+				delete(labels, k)
+			}
+		}
+		err := r.Patch(ctx, &node, client.MergeFrom(nodeCopy))
```
pcolledg-xilinx, line 147: Does `r.List()` return all nodes in the cluster? There may be thousands of nodes so we should only touch what we must: `r.Patch()` should only run if above `.hasPrefix()` matched, and potentially also only do DeepCopy if match, depending on how slow DeepCopy is.

controllers/onload_controller.go
```diff
@@ -108,6 +122,84 @@ func (r *OnloadReconciler) Reconcile(
...
+	// log.Info("Num nodes to label", "len", len(nodesToLabel))
+
+	for _, node := range nodesToLabel {
+		log.Info("node to label", "name", node.ObjectMeta.Name)
+		nodeCopy := node.DeepCopy()
+		labelKey := getLabelName(*onload)
+		node.Labels[labelKey] = onload.Spec.Onload.Version
+		err := r.Patch(ctx, &node, client.MergeFrom(nodeCopy))
+		if err != nil {
+			log.Error(err, "Failed to patch node with new label")
```
pcolledg-xilinx, line 195: I know node name is mentioned above, but I'd put it in the error too.

controllers/onload_controller.go
tcrawley-xilinx: yeah, I think there is a bug here anyway so I will update. This should have a selector to only get the nodes with that match the selector (similar to [labelNodes](https://github.com/Xilinx-CNS/kubernetes-onload/pull/52/files/f2ab2a8a54200d0948f2f1f2bf690d1e91c34b2d#diff-24b008afa2ecacf8904ab1d398c873e662b7bb8ac5ee5bf343abf7d272668c81R160-R161))

tcrawley-xilinx:
> Recommend putting funcs in dedicated `node_labels.go`.

Sure

> Doesn't this cause an overwrite battle?

No, the controller will only update the node if there isn't already a label with `onload.amd.com` ([link](https://github.com/Xilinx-CNS/kubernetes-onload/blob/f2ab2a8a54200d0948f2f1f2bf690d1e91c34b2d/controllers/onload_controller.go#L174))

> Probably better to ignore a node if any different onload.amd.com labels already exist?

That is what is already happening.

ivatet-amd:
Before committing to that approach, have you tried the inter-pod anti-affinity scheduling feature of K8s? https://github.com/kubernetes/design-proposals-archive/blob/main/scheduling/podaffinity.md and https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity.

Also, if the controller doesn't prevent the overlapping Onload CRs, what consequences does it have? The module won't load (because `--first-time`), and the device plugin won't be able to register its API with Kubelet. The only problem I see is the init container overwriting the host files, but we can try addressing it without re-labelling. It is a minor problem to address, and automatic re-labelling is a complicated feature that can go wrong in various yet unexplored ways.

Please also find a few questions I posted last week in Jira: [ON-15173](https://jira.xilinx.com/browse/ON-15173?focusedCommentId=6750622&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6750622).

PR #56, opened 2023-09-15
ON-15122: Add upgrade logic for Onload CRs

tcrawley-xilinx:
### [Patch 1/2] Cleanup: Bump KMM version to 1.1
KMM version 1.1 provides two useful features: removal of in-tree modules
and ordered upgrade of kernel modules.

Also ran `go mod tidy`
### [Patch 2/2] ON-15122: Add upgrade logic for Onload CRs
This patch adds logic to the operator so that it can upgrade (or
downgrade) the version of onload used in the cluster.

The upgrade process is designed to work with KMM's ordered upgrade
process for kernel modules, but with additions so that it can control
the lifetimes of the other operands. Upgrades take place in a "rolling"
manner, the operator will follow the upgrade procedure for one node
before starting on the next node.

Labels are used to determine the lifetime of the operand's pod on the
node, currently the operator uses two labels: one for the module
lifetime and one for the device plugin daemonset. This does expose some
of the "innards" of the operator, but it should also help recovery if
the controller crashes.

There are some other changes in this patch to the length of the
controller's reconciliation loop. Previously the controller would try to
create each operand sequentially and only exit if everything was done or
if there was an error. However, due to the amount of time that an
upgrade can take I didn't think an approach using "long reconciliations"
would be practical. So I have changed the controller so that it will
requeue a reconciliation each time the it would change the state of the
cluster.

(Apologies for the monolithic commit)

### [Patch 3/2] ON-15122: Cleanup: Refactor code for better testability

Moves logic for handling the upgrades on node (labels, pod eviction)
into a separate function which is then called from handleUpdate().

Also simplified getPodsUsingOnload to call getPodsOnNode rather than
having a separate List API call.

Both changes were made in anticipation of new unit tests and to help get
better test coverage.

### [Patch 4/2] ON-15122: Add unit tests for upgrade logic

This change adds a new set of unit tests to test the upgrade logic. In
addition to the existing tests that use EnvTest to provide the API
server, I have added gomock as new dependency for this project.

Without any mocking of the controller-runtime client we would be unable
to properly test the controller as it has to manage labels on nodes and
lifetimes of pods, and using EnvTest alone we are unable to test this
since it doesn't manage Nodes or other controllers.

The mocks for the controller-runtime client interface were automatically
generated and added as a new file in mocks/client/mock_client.go.

Unit tests that use gomock were added for functions with low code
coverage, and are not yet exhaustive of all code paths.

## Testing done
Tested manually in cluster

#### Unit tests
All unit tests pass.
```
?       github.com/Xilinx-CNS/kubernetes-onload/cmd/deviceplugin        [no test files]
?       github.com/Xilinx-CNS/kubernetes-onload/cmd/worker      [no test files]
?       github.com/Xilinx-CNS/kubernetes-onload/mocks/client    [no test files]
?       github.com/Xilinx-CNS/kubernetes-onload/pkg/deviceplugin        [no test files]
ok      github.com/Xilinx-CNS/kubernetes-onload/controllers     8.365s  coverage: 72.8% of statements
ok      github.com/Xilinx-CNS/kubernetes-onload/pkg/client_helper       5.944s  coverage: 77.8% of statements
ok      github.com/Xilinx-CNS/kubernetes-onload/pkg/control_plane       0.014s  coverage: 65.0% of statements
```


controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+			// Check that there isn't a lingering module pod
+			// This can be removed, but without it the upgrade process is more
+			// concurrent (rather than rolling), which leads to more pod
+			// restarts/failures. Both methods work, but this approach maintains
+			// the rolling nature of the update (as much as is possible).
+			pods, err := r.getPodsOnNode(ctx, kmmPodLabelSet, node.Name)
+			if err != nil {
+				return nil, err
+			} else if len(pods) > 0 {
+				log.Info("lingering module pod on node " + node.Name)
```
ivatet-amd, line 236:
```suggestion
				log.Info("Lingering module pod on node " + node.Name)
```

Also, whenever we return from here with the requeue request, we cause the "Labelled nodes with kmm label" log message, which isn't true. I haven't thought about how to address it other than returning `changesMade` .

controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+	devicePlugin.ObjectMeta.Labels[onloadVersionLabel] = onload.Spec.Onload.Version
+	devicePlugin.Spec.Template.Spec.InitContainers[0].Image = onload.Spec.Onload.UserImage
+	err = r.Patch(ctx, devicePlugin, client.MergeFrom(oldDevicePlugin))
+	if err != nil {
+		log.Error(err, "Failed to patch device plugin DaemonSet",
+			"device plugin", devicePlugin)
+		return nil, err
+	}
+
+	log.Info("Patched device plugin Daemonset", "device plugin", devicePlugin)
```
ivatet-amd, line 352:
```suggestion
	log.Info("Patched Device Plugin DaemonSet", "device plugin", devicePlugin)
```

controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+		delete(node.Labels, labelKey)
+		err := r.Patch(ctx, &node, client.MergeFrom(nodeCopy))
+		if err != nil {
+			log.Error(err, "Failed to patch node"+node.Name)
+			return err
+		}
+
+	}
+
+	labelKey = onloadLabelName(namespacedName.Name, namespacedName.Namespace)
```
ivatet-amd, line 191: Shall we make `labelKey` a parameter and shorten the function as currently there are two identical blocks of code?

controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+
+	var res *ctrl.Result = nil
+	if changesMade {
+		res = &ctrl.Result{Requeue: true}
+	}
+
+	return res, nil
+}
+
+func (r *OnloadReconciler) getBadNodes(ctx context.Context, onload *onloadv1alpha1.Onload) ([]corev1.Node, error) {
```
ivatet-amd, line 299: Shall we call "bad" nodes something else, like "nodesToUpgrade" (though it's still not the most accurate name)?

controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+	res, err = r.evictOnloadedPods(ctx, node)
+	if err != nil || res != nil {
+		return res, err
+	}
+
+	// Remove kmm label
+	oldNode := node.DeepCopy()
+	labelName := kmmLabelName(onload.Name, onload.Namespace)
+	delete(node.Labels, labelName)
+	err = r.Patch(ctx, &node, client.MergeFrom(oldNode))
```
ivatet-amd, line 481: Do you think the label deletion (`DeepCopy()`, `delete()`, `Patch()`) deserves a subroutine in the "client_helper" package?

controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+			if node.Name < min.Name {
+				min = node
+			}
+		}
+		return min
+	}(badNodes)
+
+	log.Info("Updating onload version on node "+node.Name, "onload", onload)
+
+	// Remove the onload label from the node
```
ivatet-amd, line 440: Also, I think it's worth adding a log entry indicating progress, e.g. how many nodes are on the list to upgrade, what do you think?

controllers/onload_controller.go
```diff
@@ -78,34 +90,515 @@ func (r *OnloadReconciler) Reconcile(
...
+	changesMade := false
+
+	for _, pod := range onloadedPods {
+		if pod.GetDeletionTimestamp() != nil {
+			// Pod is already being terminated, we can just continue
+			continue
+		}
+		err := r.SubResource("eviction").Create(ctx, &pod, &policyv1.Eviction{})
+		if err != nil {
+			log.Error(err, "Could not create eviction for pod: "+pod.Name)
```
ivatet-amd, line 580:
```suggestion
			log.Error(err, "Could not create eviction", "pod", pod.Name)
```

ivatet-amd (commented):
I think I've got the idea: with "mini" reconciliations, we make the Onload state closer to where we want it to be during the upgrade. I'm glad we don't need to delete DaemonSet and Modules, i.e. patching is sufficient to kick off the corresponding resource upgrade.

tcrawley-xilinx:
> Currently there is one sequential Reconcile() which is triggered by all watches yet handles each type of event in fairly cleanly encapsulated sections. Assigning watches more specifically could allow reconcilers to be called only when its domain (eg. a label selector) is changed, rather than checking everything every time.

My understanding is that for a single controller there can only be a single reconciler. Switching to a multi-controller operator could simplify the scope of the reconciler, but it would be a major architectural change and should be discussed / evaluated another time.

controllers/onload_controller.go
tcrawley-xilinx: I did have some logging (e.g. "There are N Nodes with a bad onload version"), but since it would be logged every time there was a version-mismatch it would end up spamming the log.
I'll look into updating the status to reflect the upgrade process

tcrawley-xilinx:
^ Adds unit tests for the reconciler (including tests using a `gomock`ed controller-runtime client).
Please see the top-level comment for patches and test results

controllers/onload_controller_test.go
```diff
@@ -64,6 +68,238 @@ var _ = Describe("Testing createModule function", func() {
...
+
+		mockClient = mock_client.NewMockClient(mockCtrl)
+		mockSubResourceClient = mock_client.NewMockSubResourceClient(mockCtrl)
+
+		r = &OnloadReconciler{
+			Client: mockClient,
+		}
+	})
+
+	It("should evict pods using onload", func() {
```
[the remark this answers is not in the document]
tcrawley-xilinx: Sure, I just got a bit lazy with the capitalisation.
Although, I think that the string should start with a lowercase letter ([which is what ginkgo does in the examples](https://onsi.github.io/ginkgo/#spec-subjects-it))

controllers/onload_controller.go
```diff
@@ -78,34 +92,499 @@ func (r *OnloadReconciler) Reconcile(
...
+	log := log.FromContext(ctx)
+
+	deleteLabel := func(label string) error {
+		nodes, err := r.listNodesWithLabels(ctx, label)
+		if err != nil {
+			return err
+		}
+
+		for _, node := range nodes.Items {
+			nodeCopy := node.DeepCopy()
```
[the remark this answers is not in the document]
tcrawley-xilinx: This function only operates on all nodes, whereas the removal for the upgrade process only happens on a single node.
I had a go at implementing a `deleteLabelFromNode` function that could be used in all instances, however the patch actually ended with more LoC added than removed so I'm not sure if it is worthwhile.

controllers/onload_controller.go
```diff
@@ -78,34 +92,499 @@ func (r *OnloadReconciler) Reconcile(
...
+func (r *OnloadReconciler) evictOnloadedPods(ctx context.Context, node corev1.Node) (*ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
+	onloadedPods, err := r.getPodsUsingOnload(ctx, node)
+	if err != nil {
+		log.Error(err, "Failed to get Pods using Onload")
+		return nil, err
+	}
+
+	if len(onloadedPods) == 0 {
```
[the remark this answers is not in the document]
tcrawley-xilinx: I did have logging there before, but it ended up filling the log due to the multiple short reconciliations approach.
I think that the best approach is to only log when something changes.

tcrawley-xilinx:
^ Added extra logic to remove stale labels from node.

## Testing done
#### Manual testing in the cluster:
I would either edit the selector of onload or alter the labels on the node, then would verify that the operator would remove the onload/kmm labels.

#### Automated testing
Added a new unit to cover this function.
```
$ make test
?       github.com/Xilinx-CNS/kubernetes-onload/cmd/deviceplugin        [no test files]
?       github.com/Xilinx-CNS/kubernetes-onload/cmd/worker      [no test files]
?       github.com/Xilinx-CNS/kubernetes-onload/mocks/client    [no test files]
?       github.com/Xilinx-CNS/kubernetes-onload/pkg/deviceplugin        [no test files]
ok      github.com/Xilinx-CNS/kubernetes-onload/controllers     7.282s  coverage: 73.0% of statements
ok      github.com/Xilinx-CNS/kubernetes-onload/pkg/client_helper       5.777s  coverage: 77.8% of statements
ok      github.com/Xilinx-CNS/kubernetes-onload/pkg/control_plane       0.013s  coverage: 65.0% of statements
```

ivatet-amd (approved):
Looks good to me. 🚀 

Also, thank you for the detailed commit messages.

I only suggest a trivial fix in the `ON-15122: Add upgrade logic for Onload CRs` patch:
```diff
-requeue a reconciliation each time the it would change the state of the
+requeue a reconciliation each time it would change the state of the
```

PR #57, opened 2023-09-19
ON-15217: Update Onload CRD schema

ivatet-amd:
There are two patches:

- cleanup: Remove build step from Onload CRD schema
- ON-15217: Add SFC support to Onload CRD schema

@tcrawley-xilinx has suggested the former cleanup patch since we are not planning to support the build step with the operator controller soon. Also, I have updated auto-generated files with `make generate`.

Logistics-wise, I want to rebase it on top of https://github.com/Xilinx-CNS/kubernetes-onload/pull/56 when it has merged and continue working on the SFC module support in another PR.

Testing:

```console
$ make docker-build docker-push
$ make deploy
$ oc apply -k config/samples
```


api/v1alpha1/onload_types.go
```diff
@@ -14,21 +14,34 @@ import (
...
+	OnloadBuild *OnloadKernelBuildSpec `json:"onloadBuild"`
+
+	// +optional
+	// NICModuleImage contains the out-of-tree NIC driver. If not present,
+	// the controller won't deploy the Module kind for NIC.
+	NICModuleImage string `json:"nicModuleImage"`
+
+	// +optional
+	// NICBuild is similar to the OnloadBuild field above.
+	NICBuild *NICKernelBuildSpec `json:"nicBuild"`
```
tcrawley-xilinx, line 44: If these are optional they should be probably be pointers and have `omitempty` in the json struct tags.
(I realise that this isn't consistent throughout the file, so it might be a good time to update it)

https://book.kubebuilder.io/cronjob-tutorial/api-design.html?highlight=omitempty#designing-an-api

tcrawley-xilinx (commented):
If we're going for generic nic family support I think I would prefer a term like "driver" or "netdriver" since that is what the module actually is. (apologies for bike-shedding)

api/v1alpha1/onload_types.go
ivatet-amd: Thanks, I've completely forgotten about `omitempty`. Added to the existing `OnloadKernelBuildSpec` and the new `SFCSpec` objects whose absence implies that they are empty.

Not trying to apply to the non-object/scalar fields to avoid confusion because `omitempty` doesn't distinguish between `nil` and 0.

config/crd/bases/onload.amd.com_onloads.yaml
```diff
@@ -75,23 +65,20 @@ spec:
...
                         regexp:
                           description: Regexp is a regular expression that is used
                             to match against the kernel versions of the nodes in the
                             cluster
                           type: string
+                        sfc:
+                          description: SFC specifies the driver image and various
+                            configuration knobs. If absent, the controller won't manage
+                            SFC in the given context.
+                          type: object
```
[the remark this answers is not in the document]
ivatet-amd: True, there's no need for us to comment on the content of this yet-empty type.

I'm applying your suggestion nearly as-is:

1. In "api/v1alpha1/onload_types.go" instead of this YAML, because YAML is auto-generated.
2. SFC in all capitals because it comments on the struct field and that way it is aligned with the rest of the file.

Hope that's OK.

ivatet-amd:
Rebased, retested and updated PR.

PR #58, opened 2023-09-21
ON-15217: Create Module kind for SFC kernel module

ivatet-amd:
The Onload operator controller will create the SFC module kind if the SFC field is present in the Onload CR. In this case, the SFC module kind will reuse the Onload's kernel image.

The controller also takes care of the upgrade of the SFC module kind along with Onload. Moreover, if the old Onload CR does not contain SFC but the new Onload CR does, it will create the new module kind; if vice-versa, it will delete the SFC module kind.

Reviewed-by: Peter Colledge <peter.colledge@amd.com>
Reviewed-by: Thomas Crawley <thomas.crawley@amd.com>

## Testing

I used the image from `v3.0.0-pre` with minor changes to include the SFC and driver link kernel modules:

```diff
diff --git a/onload/kmm/Dockerfile b/onload/kmm/Dockerfile
index 8716674..8697799 100644
--- a/onload/kmm/Dockerfile
+++ b/onload/kmm/Dockerfile
@@ -13,17 +13,26 @@ RUN mkdir -p /build/onload
 RUN tar xzf onload.tar.gz -C /build/onload --strip-components=1
 WORKDIR /build/onload/

+# Onload
 RUN scripts/onload_build --kernel --kernelver ${KERNEL_VERSION} ${ONLOAD_BUILD_PARAMS}
 RUN scripts/onload_install --nobuild --kernelfiles --kernelver ${KERNEL_VERSION}

+# SFC
+RUN make -j8 -C src/driver/linux_net CONFIG_SFC_VDPA= CONFIG_SFC_MTD= KVER=${KERNEL_VERSION}
+
 FROM ubi8:8.8
 ARG KERNEL_VERSION

 RUN yum -y install kmod && yum clean all
 RUN mkdir -p /opt/lib/modules/${KERNEL_VERSION}

+# Onload
 COPY --from=builder /lib/modules/${KERNEL_VERSION}/extra/onload.* /opt/lib/modules/${KERNEL_VERSION}
 COPY --from=builder /lib/modules/${KERNEL_VERSION}/extra/sfc_char.* /opt/lib/modules/${KERNEL_VERSION}
 COPY --from=builder /lib/modules/${KERNEL_VERSION}/extra/sfc_resource.* /opt/lib/modules/${KERNEL_VERSION}

+# SFC
+COPY --from=builder /build/onload/src/driver/linux_net/drivers/net/ethernet/sfc/sfc.ko /opt/lib/modules/${KERNEL_VERSION}/
+COPY --from=builder /build/onload/src/driver/linux_net/drivers/net/ethernet/sfc/sfc_driverlink.ko /opt/lib/modules/${KERNEL_VERSION}/
+
 RUN /usr/sbin/depmod -b /opt
```

A list of operands in runtime when `sfc` is present in the Onload CR:

```console
$ oc get pod
NAME                                          READY   STATUS      RESTARTS        AGE
onload-sample-onload-device-plugin-ds-j2qdz   2/2     Running     2 (7m10s ago)   9m29s
onload-sample-onload-device-plugin-ds-rb79p   2/2     Running     3 (7m4s ago)    9m29s
onload-sample-onload-module-9w67l-g2td5       1/1     Running     0               9m29s
onload-sample-onload-module-9w67l-t96cz       1/1     Running     0               9m29s
onload-sample-sfc-module-x7jnc-2wfvf          1/1     Running     0               9m29s
onload-sample-sfc-module-x7jnc-k69rp          1/1     Running     0               9m29s
```

The controller won't create the SFC Module kind if `sfc` is not present, and it's also covered with a unit test.

Also, due to the racing nature when the controller adds operands simultaneously, this can trigger the Device Plugin (expected) restart when it discovers changes in the list of NICs.:
```
F0920 15:58:47.133869 1999974 manager.go:92] SFC interfaces on host have changed ([0] -> [sf0])
```
(https://github.com/Xilinx-CNS/kubernetes-onload/blob/57f709d48a01d5e1a40483205d2e2bee9277942b/pkg/deviceplugin/manager.go#L92C3-L92C3)

### DIFF
--- a/controllers/onload_controller.go
+++ b/controllers/onload_controller.go
@@ -5,6 +5,7 @@ package controllers
 import (
 	"cmp"
 	"context"
+	"fmt"
 	"slices"
 	"time"
 
@@ -106,7 +107,7 @@ func (r *OnloadReconciler) Reconcile(
 		log.Info("Error creating module")
 		return ctrl.Result{}, err
 	} else if res != nil {
-		log.Info("Added Onload Module")
+		log.Info("Added Module(s)")
 		return *res, nil
 	}
 
@@ -143,9 +144,14 @@ func (r *OnloadReconciler) Reconcile(
 
 const kmmModuleLabelPrefix = "kmm.node.kubernetes.io/version-module"
 const onloadModuleNameSuffix = "-onload-module"
+const sfcModuleNameSuffix = "-sfc-module"
 
-func kmmLabelName(name, namespace string) string {
+func kmmOnloadLabelName(name, namespace string) string {
 	return kmmModuleLabelPrefix + "." + namespace + "." + name + onloadModuleNameSuffix
+}
+
+func kmmSFCLabelName(name, namespace string) string {
+	return kmmModuleLabelPrefix + "." + namespace + "." + name + sfcModuleNameSuffix
 }
 
 const onloadLabelPrefix = "onload.amd.com/"
@@ -194,7 +200,12 @@ func (r *OnloadReconciler) deleteLabels(ctx context.Context, namespacedName type
 		return nil
 	}
 
-	err := deleteLabel(kmmLabelName(namespacedName.Name, namespacedName.Namespace))
+	err := deleteLabel(kmmOnloadLabelName(namespacedName.Name, namespacedName.Namespace))
+	if err != nil {
+		return err
+	}
+
+	err = deleteLabel(kmmSFCLabelName(namespacedName.Name, namespacedName.Namespace))
 	if err != nil {
 		return err
 	}
@@ -209,41 +220,73 @@ func (r *OnloadReconciler) deleteLabels(ctx context.Context, namespacedName type
 
 func (r *OnloadReconciler) addKmmLabelsToNodes(ctx context.Context, onload *onloadv1alpha1.Onload) (*ctrl.Result, error) {
 	log := log.FromContext(ctx)
-	kmmPodLabelSet := labels.Set{"kmm.node.kubernetes.io/module.name": onload.Name + onloadModuleNameSuffix}
 
-	labels := labels.FormatLabels(onload.Spec.Selector)
-	labelKey := kmmLabelName(onload.Name, onload.Namespace)
-
-	nodes, err := r.listNodesWithLabels(ctx, labels)
+	// Figure out a list of nodes that match Onload's selector.
+	onloadLabels := labels.FormatLabels(onload.Spec.Selector)
+	nodes, err := r.listNodesWithLabels(ctx, onloadLabels)
 	if err != nil {
 		return nil, err
 	}
 
 	changesMade := false
 
-	for _, node := range nodes.Items {
-		if _, found := node.Labels[labelKey]; !found {
-			// Check that there isn't a lingering module pod.
-			// This can be removed, but without it the upgrade process is more
-			// concurrent (rather than rolling), which leads to more pod
-			// restarts/failures. Both methods work, but this approach maintains
-			// the rolling nature of the update (as much as is possible).
-			pods, err := r.getPodsOnNode(ctx, kmmPodLabelSet, node.Name)
-			if err != nil {
-				return nil, err
-			} else if len(pods) > 0 {
-				log.Info("Lingering Module Pod(s)", "Node", node.Name)
-				return &ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
-			}
+	addKmmLabelToNode := func(node corev1.Node, labelKey string, podLabelSet labels.Set,
+	) (*ctrl.Result, error) {
+		if _, found := node.Labels[labelKey]; found {
+			return nil, nil
+		}
 
-			nodeCopy := node.DeepCopy()
-			node.Labels[labelKey] = onload.Spec.Onload.Version
-			err = r.Patch(ctx, &node, client.MergeFrom(nodeCopy))
-			if err != nil {
-				log.Error(err, "Failed to patch Node with new label")
-				return nil, err
+		// We can only add the SFC label if the Onload label matches
+		// the Onload CR version. Otherwise, we are likely within the
+		// upgrade workflow, where we soon delete old KMM labels and
+		// add new ones, causing the SFC kernel module to reload
+		// unnecessarily.
+		if labelKey != kmmOnloadLabelName(onload.Name, onload.Namespace) &&
+			node.Labels[kmmOnloadLabelName(onload.Name, onload.Namespace)] != onload.Spec.Onload.Version {
+			return nil, nil
+		}
+
+		// Check that there isn't a lingering module pod.
+		// This can be removed, but without it the upgrade process is more
+		// concurrent (rather than rolling), which leads to more pod
+		// restarts/failures. Both methods work, but this approach maintains
+		// the rolling nature of the update (as much as is possible).
+		pods, err := r.getPodsOnNode(ctx, podLabelSet, node.Name)
+		if err != nil {
+			return nil, err
+		} else if len(pods) > 0 {
+			log.Info("Lingering Module Pod(s)", "Node", node.Name)
+			return &ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
+		}
+
+		nodeCopy := node.DeepCopy()
+		node.Labels[labelKey] = onload.Spec.Onload.Version
+		err = r.Patch(ctx, &node, client.MergeFrom(nodeCopy))
+		if err != nil {
+			log.Error(err, "Failed to patch Node with new label")
+			return nil, err
+		}
+		changesMade = true
+		return nil, nil
+	}
+
+	for _, node := range nodes.Items {
+		// Try add Onload labels
+		res, err := addKmmLabelToNode(node,
+			kmmOnloadLabelName(onload.Name, onload.Namespace),
+			labels.Set{"kmm.node.kubernetes.io/module.name": onload.Name + onloadModuleNameSuffix})
+		if res != nil || err != nil {
+			return res, err
+		}
+
+		// Add SFC labels conditionally
+		if onloadUsesSFC(onload) {
+			res, err := addKmmLabelToNode(node,
+				kmmSFCLabelName(onload.Name, onload.Namespace),
+				labels.Set{"kmm.node.kubernetes.io/module.name": onload.Name + sfcModuleNameSuffix})
+			if res != nil || err != nil {
+				return res, err
 			}
-			changesMade = true
 		}
 	}
 
@@ -252,7 +295,13 @@ func (r *OnloadReconciler) addKmmLabelsToNodes(ctx context.Context, onload *onlo
 		return &ctrl.Result{Requeue: true}, nil
 	}
 
-	return r.removeStaleLabels(ctx, onload, labelKey)
+	// Remove stale kmm labels from nodes that don't match the module kind's selector.
+	res, err := r.removeStaleLabels(ctx, onload, kmmOnloadLabelName(onload.Name, onload.Namespace))
+	if res != nil || err != nil {
+		return res, err
+	}
+
+	return r.removeStaleLabels(ctx, onload, kmmSFCLabelName(onload.Name, onload.Namespace))
 }
 
 func (r *OnloadReconciler) addOnloadLabelsToNodes(ctx context.Context, onload *onloadv1alpha1.Onload) (*ctrl.Result, error) {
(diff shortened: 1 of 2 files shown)


controllers/onload_controller.go
```diff
@@ -194,41 +199,79 @@ func (r *OnloadReconciler) deleteLabels(ctx context.Context, namespacedName type
...
+			true, // Onload module kind is always used for Onload CR
+			kmmOnloadLabelName(onload.Name, onload.Namespace),
+			labels.Set{"kmm.node.kubernetes.io/module.name": onload.Name + onloadModuleNameSuffix},
+		},
+		{
+			onloadUsesSFC(onload),
+			kmmSFCLabelName(onload.Name, onload.Namespace),
+			labels.Set{"kmm.node.kubernetes.io/module.name": onload.Name + sfcModuleNameSuffix},
+		},
+	}
```
tcrawley-xilinx, line 245: Rather than including a conditional bool in the struct, have you considered conditionally adding the the sfc label to the slice?
Something like:
```
kmmLabels = []labels{ Onload definition}
if onloadUsesSFC(onload) {
    kmmLabels = append(kmmLabels, {sfc definition})
}
```

Then you could remove the check for isUsed later.

controllers/onload_controller.go
```diff
@@ -603,23 +721,62 @@ func (r *OnloadReconciler) evictOnloadedPods(ctx context.Context, node corev1.No
 	return &ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
 }
 
+type getKernelMapping func(onloadv1alpha1.OnloadKernelMapping) *kmm.KernelMapping
+
+type moduleConfig struct {
+	modprobeArg string
+	moduleName  string
+
+	getKernelMapping getKernelMapping
```
tcrawley-xilinx, line 730: Can you make the type and field names different please? (Also applies to other uses of getKernelMapping)

controllers/onload_controller.go
```diff
@@ -603,23 +721,62 @@ func (r *OnloadReconciler) evictOnloadedPods(ctx context.Context, node corev1.No
...
 	onload *onloadv1alpha1.Onload,
 ) (*ctrl.Result, error) {
 	log := log.FromContext(ctx)
 
-	moduleNames := []string{
-		onload.Name + onloadModuleNameSuffix,
-		// onload.Name + "-sfc-module",
+	moduleConfigs := []moduleConfig{
+		{"onload", onload.Name + onloadModuleNameSuffix, getOnloadKernelMapping},
+		{"sfc", onload.Name + sfcModuleNameSuffix, getSFCKernelMapping},
```
tcrawley-xilinx, line 771: Have you thought about adding a check with `onloadUsesSFC` here to conditionally add the module config to the slice. It could simplify the for loop and `createModule` since we would only enter them when we actually want to do something and so could remove the some of the `== nil` or `len(...) == 0` checks.

tcrawley-xilinx (commented):
Looks good overall.
I've suggested changes in a couple places where i think we could simplify things, but there doesn't seem to be any logical errors.

controllers/onload_controller.go
```diff
@@ -432,6 +479,70 @@ func (r *OnloadReconciler) handleModuleUpdate(ctx context.Context, onload *onloa
...
+				if res != nil || err != nil {
+					return res, err
+				}
+			} else {
+				err = r.Delete(ctx, module)
+				return &ctrl.Result{Requeue: true}, err
+			}
+		} else if apierrors.IsNotFound(err) {
+			if moduleConfig.isUsed {
+				return &ctrl.Result{Requeue: true}, nil
```
tcrawley-xilinx, line 532: I don't think this case can happen, the reconciliation loop will check the existence of the SFC module before it reaches the upgrade code. Which I think means we can continue in either case if the module doesn't exist.

controllers/onload_controller_test.go
```diff
@@ -165,34 +245,72 @@ var _ = Describe("Testing using mocked client", func() {
...
+				List(gomock.Any(), &corev1.NodeList{}, gomock.Any()).
+				SetArg(1, nodes).
+				Return(nil).
 				Times(1)
 
+			// This can be either "list nodes" or "patch node"
+			previousCall := listNodesCall
+
+			modules := []string{"onload", "sfc"}
+			for _ = range modules {
```
tcrawley-xilinx, line 281:
```suggestion
			for range modules {
```
Since you're not using either the index or the value, you can skip the dummy definition

controllers/onload_controller.go
```diff
@@ -603,23 +721,62 @@ func (r *OnloadReconciler) evictOnloadedPods(ctx context.Context, node corev1.No
...
+}
+
+// Return true if any of the kernel mappings contain a non-nil SFC field.
+func onloadUsesSFC(onload *onloadv1alpha1.Onload) bool {
+	for _, kmapSpec := range onload.Spec.Onload.KernelMappings {
+		kmap := getSFCKernelMapping(kmapSpec)
+		if kmap != nil {
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
ivatet-amd: That's fantastic, thanks. I wonder if there are more occurrences where `slices` could help.

controllers/onload_controller.go
ivatet-amd: Totally. 👍

controllers/onload_controller.go
ivatet-amd: Yes, it's a good idea. Also, given that it means no changes in `createModule()`, we could remove a new unit test for `== nil` and `len(...) == 0` without sacrificing the test coverage because the optional SFC module kind creation is tested with EnvTests.

controllers/onload_controller.go
ivatet-amd: It's a good idea to disambiguate it.

How about these (in the latest revision):

- Function type is `onload2kmmKernelMapper`.
- Field is `getKernelMap`.
- Onload and SFC functions are `onloadKernelMapper` and `sfcKernelMapper`.

However, naming is a hard problem! I've noticed https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch uses the `MutateFn` type. Do you think the `Fn` suffix would improve readability in our case?

controllers/onload_controller.go
ivatet-amd: Yes, well spotted. How about raising an error instead because this situation should not occur, as you outline? I believe we'll re-enter reconciliation after that, so it's still not the same as `continue`.

controllers/onload_controller_test.go
ivatet-amd: Neat.

ivatet-amd:
@tcrawley-xilinx, thanks for the prompt review.

I've smoke-tested some scenarios, but still want to test it more.

Interestingly, have also discovered a flaky unit test failure with Module in EnvTest:
```
  [FAILED] Expected
      <string>:
  to equal
      <string>: kernel-image:tag2
  In [It] at: /home/iteterev/lab/kubernetes-onload/controllers/onload_controller_test.go:718 @ 10/02/23 11:29:07.373
```
Sometimes, the module is found, but the container image is empty. I'll take a closer look too.

controllers/onload_controller.go
ivatet-amd: Re-opening, because there's another similar `isUsed` occurrence in `handleModuleUpdate()`.

controllers/onload_controller.go
tcrawley-xilinx: I'd be happy with a type name of something like `kernelMapperFn`

controllers/onload_controller.go
ivatet-amd: Yes, this `isUsed` would be non-trivial to remove.

controllers/onload_controller.go
```diff
@@ -432,6 +479,73 @@ func (r *OnloadReconciler) handleModuleUpdate(ctx context.Context, onload *onloa
...
+		isUsed     bool
+		moduleName string
+
+		getKernelMap onload2kmmKernelMapper
+	}{
+		{
+			true, // Onload module kind is always used for Onload CR
+			onload.Name + onloadModuleNameSuffix,
+			onloadKernelMapper,
+		},
```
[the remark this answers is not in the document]
ivatet-amd: It's a good idea, thanks.

controllers/onload_controller.go
```diff
@@ -209,41 +220,73 @@ func (r *OnloadReconciler) deleteLabels(ctx context.Context, namespacedName type
...
+	onloadLabels := labels.FormatLabels(onload.Spec.Selector)
+	nodes, err := r.listNodesWithLabels(ctx, onloadLabels)
 	if err != nil {
 		return nil, err
 	}
 
 	changesMade := false
 
+	addKmmLabelsToNode := func(node corev1.Node, labelKey string, podLabelSet labels.Set,
+	) (*ctrl.Result, error) {
```
tcrawley-xilinx, line 234: You can remove `*ctrl.Result` from the return types as you never return a non-nil value. Also, could you return whether or not the function made a change, so that you are modifying a variable defined outside of the function (I know that this version works, its just a bit ugly).

The return type could be something like `(bool, error)`

controllers/onload_controller.go
```diff
@@ -470,11 +564,14 @@ func (r *OnloadReconciler) handleNodeUpdate(ctx context.Context, onload *onloadv
...
-	// Remove kmm label
-	labelName := kmmLabelName(onload.Name, onload.Namespace)
-	err = r.deleteLabelFromNode(ctx, node, labelName)
+	// Remove kmm labels
+	err = r.deleteLabelFromNode(ctx, node, kmmOnloadLabelName(onload.Name, onload.Namespace))
+	if err != nil {
+		return nil, err
+	}
+
+	err = r.deleteLabelFromNode(ctx, node, kmmSFCLabelName(onload.Name, onload.Namespace))
```
tcrawley-xilinx, line 573: You should try to remove the sfc module first, then the onload label.
If there is an error here when the reconciler is requeued it won't try to remove the lingering sfc kmm label. (The onload kmm label is used to check whether the onload version on a node is correct).

tcrawley-xilinx (changes requested):

controllers/onload_controller.go
ivatet-amd: > You can remove *ctrl.Result from the return types as you never return a non-nil value. 

It doesn't sound right because the nested function requests reconciliation if there are running Onloaded pods ("Lingering Module Pod(s)").

> its just a bit ugly

This ugliness is somewhat expected here because the function itself is non-trivial and tries to batch the reconciliations if there are any changes. (I think about the boolean as a deferred requeue/reconciliation request.) If anyone trips over this boolean, they will notice the batching quicker, what do you reckon?

If you still don't like it, maybe we should address the root cause of it and requeue without batching?

controllers/onload_controller.go
ivatet-amd: Well spotted!

I've added a comment too.

controllers/onload_controller.go
tcrawley-xilinx: Fair enough, leave as is then

ivatet-amd:
In the latest revision:

- Added `InTreeModuleToRemove` for the SFC module kind.
- Used `DescribeTable()` for the Onload upgrade unit test matrix.

Credits to @tcrawley-xilinx.

Testing in the cluster is pending.

ivatet-amd:
Re on a flaky unit test failure with Module in EnvTest.

It has been reproduced in GH CI for the previous revision of the PR. We believe it's because the asynchronous `Eventually()` is racing with the Onload upgrade and potentially checking the module's state while the controller is patching the module's spec. Hence the test is operating on the outdated (wrong) module's state and fails depending on timings. Tightening the condition in `Eventually()` should address this issue.

The same applies to the earlier introduced Onload Device Plugin checks that ensure the image init container's image matches the upgraded Onload image.

tcrawley-xilinx (approved):

ivatet-amd:
Tested in a dev. cluster.

Test plan:
1. Create Onload CR 8.1.0 without SFC.
2. Update Onload CR to 8.1.1 without SFC.
3. Downgrade back to 8.1.0.
4. Upgrade Onload CR to 8.1.1 with SFC.
5. Check I/O acceleration with `sfnt-pingpong` and leave the pods running.
6. Downgrade back to 8.1.0.

Things to check:
* At each upgrade/downgrade, check `lsmod` periodically. The SFC and Onload kernel modules all get unloaded and then, shortly after, reloaded one time during the module kinds upgrade.
* The SFC module kind and the kernel module are only present at steps 4 and 5 above.
* The `sfnt-pingpong` I/O at step 5 is accelerated.
* The Onloaded `sfnt-pingpong` pods are evicted at step 6 (and actually deleted, which must be root-caused by not running these as part of the deployment and specifying `nodeName`).

PR #63, opened 2023-10-06
ON-15296: Add unit test for conditional node labelling

ivatet-amd:
The `addKmmLabelsToNodes()` method should not add the SFC KMM label to the node with the Onload KMM label with the value not matching the requested Onload CR's version. This is to not interfere with the upgrade workflow.


ivatet-amd:
It's a bit of a corner case that this PR is trying to unit-test!

PR #65, opened 2023-10-12
cleanup: Dependabot config

pcolledg-amd:
[`commit-message`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file?learn=dependency_version_updates&learnProduct=code-security#commit-message)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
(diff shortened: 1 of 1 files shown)


tcrawley-xilinx (approved):
Looks good.
I suppose this could cause issue with the ContainerFileNotice files being out of sync, but that is only something we would need to worry about for releases.

PR #69, opened 2023-11-08
ON-15183: Update README.md

ivatet-amd:


README.md
```diff
@@ -1,75 +1,138 @@
...
+### Prepare Kubernetes cluster
 
-It uses [Controllers](https://kubernetes.io/docs/concepts/architecture/controller/),
-which provide a reconcile function responsible for synchronizing resources until the desired state is reached on the cluster.
+The Onload Operator v3 depends on third-party software:
 
-### Test It Out
-1. Install the CRDs into the cluster:
+1. Kernel Module Manamagent (KMM) Operator v1.1.1.
+2. CNI that supports ipvlan or macvlan configurations. The Onload Operator currently suggests Multus CNI. A sample configuration is https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/examples/macvlan-pod.yml.
```
tcrawley-xilinx, line 53: I don't think we are actually tied to `ipvlan` or `macvlan`. We should be able to use multus in general (although we have only tested with ipvlan/macvlan)
```suggestion
2. Multus CNI. A sample configuration using macvlan is https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/examples/macvlan-pod.yml.
```

README.md
```diff
@@ -1,75 +1,138 @@
...
+## Run Onloaded application
+
+At this point, Onload is deployed at the cluster, and the users can run Onloaded applications, e.g.
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+  annotations:
+    k8s.v1.cni.cncf.io/networks: ipvlan-sf0
```
tcrawley-xilinx, line 108: Could you add a note/comment for this line saying that this should be the name of the NetworkAttachmentDefinition that they created earlier (and perhaps make it obvious that they need to create the NetworkAttachmentDefinition, just having the CNI isn't sufficient).

tcrawley-xilinx (commented):
Looks good, I wonder if we should change all of the examples to something more obviously fake like `example.com`.

README.md
ivatet-amd: Thanks, applied. Also, added a note that `NetworkAttachmentDefinition` is used later.

PR #82, opened 2023-11-21
ON-15420: Support in-cluster builds for onload-module image

tcrawley-xilinx:
### [1/3] ON-15420: Add module build options to onload CRD

Adds a subset of the KMM build options to the OnloadKernelMapping
struct. If present these options are converted in a KMM compatible form
and added to the Module struct. This allows the KMM Operator to perform
an in-cluster build.

The build field is cleared for the sfc module as otherwise two identical
builds would start on the cluster which would both be trying to write
the same image. It safe to clear the sfc definition and leave the onload
module untouched as it is not currently possible for the operator to
create a sfc module kind without a corresponding onload module kind, and
since they use the same images it will be able to start the pod when the
image is present.

### [2/3] ON-15420: Add unit tests for *KernelMapper functions

Probably slightly overkill, but I though since I was updating the code around here I may as well create some new unit tests.

### [3/3] ON-15420: Add in-cluster build example

This commit adds an example of how to use an in-cluster build, this is
done using kustomize to patch the onload CR definition and add a build
section which points at the onload-module-dockerfile added previously.

This change necessitated moving the sample onload CR yaml file into the
new config/samples/base directory to prevent kustomize complaining about
a circular dependency.

To create the onload-sample without the in-cluster build you should run:
```
kubectl create -k config/samples
```
OR
```
kubectl create -k config/samples/base
```
and to create it with the in-cluster build you should run:
```
kubectl create -k config/samples/in-cluster-build
```

## Testing done

Unit tests pass

#### Manual testing:
Created an onload CR using the patch shown in commit 3 (slightly modified to push to the openshift internal registry), it was able to create the image then both modules could pull the newly created image. onload was able to accelerate traffic as normal.


tcrawley-xilinx:
I realised after I posted this that I didn't test upgrading onload with an in-cluster build.
Tested using
```
$ oc edit onload onload-sample --output-patch
Patch: {"spec":{"onload":{"kernelMappings":[{"build":{"buildArgs":[{"name":"ONLOAD_BUILD_PARAMS","value":""},{"name":"ONLOAD_SOURCE","value":"kitchen-sink.kube.test:5000/onload-source:6c4afc8"}],"dockerfileConfigMap":{"name":"onload-module-dockerfile-d7bbgm4fft"}},"kernelModuleImage":"image-registry.openshift-image-registry.svc:5000/onload-clusterlocal/onload-module:6c4afc8","regexp":"^.*\\.x86_64$"}],"userImage":"kitchen-sink.kube.test:5000/onload-user:6c4afc8","version":"8.1.2"}}}
```
Then monitored the state of things.

It works, but basically breaks our claim that the upgrade is "rolling". The takedown of the old versions is unaffected, but the operator doesn't wait until the build finishes/module comes up, so progresses onto the next node while it is still building. In my test this resulted it removing the old versions on all of the nodes before the build finished.
I have some ideas for how to prevent this issue, but it will (slightly) increase the complexity of the upgrade logic.

ivatet-amd (approved):
> since I was updating the code around here I may as well create some new unit tests.

"Heavy plus one".

PR #83, opened 2023-11-22
ON-14857: Add program to produce MachineConfigs

tcrawley-xilinx:
Added a small program to wrap around the mcproducer package, and provide
a basic set of command line options.

Can be run with either `go run cmd/machineconfig/main.go` or by
compiling with `go build` and subsequently ran anywhere.

## Testing done

No testing in cluster.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,13 +28,14 @@ require (
 	github.com/go-logr/zapr v1.2.4 // indirect
 	github.com/go-openapi/jsonpointer v0.19.6 // indirect
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
-	github.com/go-openapi/swag v0.22.3 // indirect
+	github.com/go-openapi/swag v0.22.4 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/kubernetes-sigs/kernel-module-management v1.1.0
 	github.com/onsi/ginkgo/v2 v2.13.1
 	github.com/onsi/gomega v1.30.0
+	github.com/rh-ecosystem-edge/kernel-module-management v0.0.0-20231024110254-591541ea5a45
```
ivatet-amd, line 10: It's a bit annoying, but could you also update ContainerFileNotice to reflect changes in go.mod?

tcrawley-xilinx:
Closing in favour of #86.
After doing some more research I don't think that `mcproducer` is the right way to go as it hardcodes the ignition version which will limit what this is compatible with.

PR #85, opened 2023-11-22
ON-15386: Support onload device plugin without automatic preload

tcrawley-xilinx:
Allows the user to disable the device plugin setting `LD_PRELOAD` in the pod, as well as adding extra options to the device plugin that allows for the mounting of the `onload` script and for configurable mount locations in the user's container. These changes will affect all pods that request an onload resource, unfortunately it was a bit too complicated to allow this configuration on a per-pod basis.

(Apologies for the large PR)

### [1/8] ON-15386: Add usePreload flag to device plugin
Adds new "usePreload" flag to the device plugin's command line options.
If set the LD_PRELOAD environment variable will be set in the user's
pod, otherwise the container's environement will be unaffected.
Also added new unit tests to cover these options.

### [2/8] ON-15386: Add preload option to DevicePlugin in the onload CRD
Adds a new options that corresponds to the flag added to the device
plugin. Also added the necessary plumbing to update the DaemonSet and
added a unit test.

### [3/8] ON-15386: Add mountOnload flag to device plugin
If set the Onload Device Plugin will mount the onload script as a file
in the container's file system. This option is mutually exclusive with
preload.

Also added unit test

### [5/8] Cleanup: Add NicManagerConfig struct
With the addition of new options to the device plugin the number of
parameters of some functions (in particular NewNicManager) was getting
unwieldy. This moves all of the options into a new struct that is added
to the NicManager struct.

### [6/8] ON-15386: Add options to configure onload file locations
Previously the location of onload files in the host and destination were
fixed constants, this isn't ideal if we want to support different
environments. This change adds new command line options to specify the
base onload location on the host as well as the destination locations
for libraries and binaries.

### [7/8] ON-15228: Add extra unit test for device plugin config
Using LD_PRELOAD and mount onload are mutually exclusive options, so we
should have a test to verify this.

## Testing done
Tested in cluster, using defaults has no difference in behaviour.
When setting:
```text
preload: false
mountOnload: true
baseMountPath: ""
```
Then running `onload ./sfnt-pingpong ...` in the example pod showed acceleration working.


controllers/onload_controller_test.go
```diff
@@ -835,6 +835,81 @@ var _ = Describe("onload controller", func() {
...
+				}, timeout, pollingInterval).Should(BeTrue())
+
+				Expect(devicePlugin.Spec.Template.Spec.Containers).Should(
+					ContainElement(MatchFields(IgnoreExtras, Fields{
+						"Args": ContainElement(Equal("-mountOnload=false")),
+					})),
+				)
+			})
+
+			It("should pass the values of filepaths through", func() {
```
ivatet-amd, line 882: Shall we bundle all three with `DescribeTable`? Because it's a multi-commit PR, one commit could introduce the table, and the follow-up ones could add more `Entry`.

pkg/deviceplugin/manager.go
```diff
@@ -40,20 +64,26 @@ func (manager *NicManager) GetDeviceFiles() []*pluginapi.DeviceSpec {
...
 	manager := &NicManager{
-		interfaces:     nics,
-		maxPodsPerNode: maxPods,
+		interfaces: nics,
+		config:     config,
 	}
 	manager.envs = make(map[string]string)
 	manager.initDevices()
+
+	if manager.config.Preload && manager.config.MountOnload {
```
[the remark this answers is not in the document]
tcrawley-xilinx: Yeah, we could probably do something with [kubebuilder validations](https://book.kubebuilder.io/reference/markers/crd-validation), but since in theory the device plugin should work without the controller we would still need a check in the device plugin.

abower-amd (approved):
Approve in principle - but requires technical review.

controllers/onload_controller_test.go
```diff
@@ -830,11 +846,35 @@ var _ = Describe("onload controller", func() {
 
 				Expect(devicePlugin.Spec.Template.Spec.Containers).Should(
 					ContainElement(MatchFields(IgnoreExtras, Fields{
-						"Args": ContainElement(Equal("-maxPods=1")),
+						"Args": BeEmpty(),
```
ivatet-amd, line 849: This check can be removed too.

controllers/onload_controller_test.go
```diff
@@ -791,36 +792,51 @@ var _ = Describe("onload controller", func() {
 			Entry("Upgrade Onload with SFC", &onloadv1alpha1.SFCSpec{}, &onloadv1alpha1.SFCSpec{}),
 		)
 
-		Describe("Testing Device Plugin options", func() {
-			It("shouldn't add anything when empty", func() {
-				devicePlugin := appsv1.DaemonSet{}
-				devicePluginName := types.NamespacedName{
-					Name:      onload.Name + "-onload-device-plugin-ds",
-					Namespace: onload.Namespace,
+		checkDevicePluginOptions := func(spec *onloadv1alpha1.DevicePluginSpec, devicePlugin *appsv1.DaemonSet) {
```
ivatet-amd, line 795: Shall we remove `checkDevicePluginOptions()`, e.g. https://github.com/Xilinx-CNS/kubernetes-onload/commit/bed2e0c9a4baf2ca45b0db5b7c2f1b0db5ce4672?

pkg/deviceplugin/manager.go
```diff
@@ -16,19 +16,43 @@ import (
...
+	SetPreload     bool
+	MountOnload    bool
+	HostPathPrefix string
+	BaseMountPath  string
+	BinMountPath   string
+	LibMountPath   string
+	NeedNic        bool
+}
+
+// Ideally this would be const, but go doesn't support const structs
```
ivatet-amd, line 31:
```suggestion
// Ideally this would be const, but go doesn't support const structs.
```

api/v1alpha1/onload_types.go
```diff
@@ -76,6 +76,7 @@ type OnloadSpec struct {
 }
 
 // Currently unimplemented
+// +kubebuilder:validation:XValidation:message="SetPreload and MountOnload mutually exclusive",rule="!(self.setPreload && self.mountOnload)"
```
ivatet-amd, line 79: That's fantastic. 👍

controllers/onload_controller.go
```diff
@@ -972,21 +1016,11 @@ func (r *OnloadReconciler) createDevicePluginDaemonSet(
 			Privileged: ptr.To(true),
 		},
 		VolumeMounts: []corev1.VolumeMount{
-			{
-				MountPath: "/opt/onload",
```
ivatet-amd, line 976: Thanks for removing it.

PR #87, opened 2023-11-24
ON-15311: Add script to generate configmap from profile

tcrawley-xilinx:
Adds a new script `process_profile.sh` which is essentially a stripped down version of the `onload` script that only includes the logic used to process the profiles. The new script has the same syntax as the original onload version, but now instead of setting environment variables it will output the yaml text definition of a ConfigMap which can be used to set environment variables in a user's pod.

Currently the script produces ConfigMaps with a fixed naming structure, for example if you want to create a ConfigMap from a profile called `name.opf` the generated name will be `onload-name-profile`.

If you want to apply the generate ConfigMap straight away, then you can simply run:
`./process_profile.sh -p /path/to/profile.opf | oc apply -f -`

This patch also includes an example latency profile in config/profiles

## Testing done

Verified the generated ConfigMap against the known good examples from `v3.0.0-pre`

Also `shellcheck` has no complaints/warnings from the new file

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,55 @@ There are two key fields in the above example CR:
 
 No further modifications are required to enable Onloaded applications.
 
+### Using Onload profiles
+
+If you want to run your onloaded application with a runtime profile we suggest
+using a ConfigMap to set the environment variables in the pod. We have included
+and example definition for the latency profile in `config/samples/profiles`
+directory. This can be applied as so:
+```text
+oc apply -k config/samples/profiles
+```
+which creates a new ConfigMap called `onload-latency-profile` in the current
+namespace.
+
+Then to use this in you pod, add the following to the container spec in your pod
+definition:
+
+```diff
+diff --git a/testpod.yaml b/testpod.yaml
+index 5259d67..ced3b2a 100644
+--- a/testpod.yaml
++++ b/testpod.yaml
+@@ -16,6 +16,9 @@ spec:
+     imagePullPolicy: Always
+     command:
+     - /test
++    envFrom:
++      - configMapRef:
++          name: onload-latency-profile
+     resources:
+       limits:
+         amd.com/onload: 1
+```
+
+#### Converting an existing profile
+If you have an existing profile defined as a `.opf` file you can generate a new
+ConfigMap definition from this using the `./scripts/profile_to_configmap.sh`
+script.
+
+`profile_to_configmap.sh` takes in a comma separated list of profiles and will
+output the text definition of the ConfigMap which can be saved into a file, or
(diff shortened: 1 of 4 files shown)


config/profiles/latency.yaml
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: onload-latency-profile
+data:
+  # SPDX-License-Identifier: MIT
```
ivatet-amd, line 8: Onload profiles have a different license: BSD-2-Clause.

Shall we also add more headers to specify the source file and the version, e.g. https://github.com/Xilinx-CNS/onload/blob/master/src/driver/linux_onload/cplane.c#L1715.

scripts/profiles/process_profile.sh
```diff
@@ -0,0 +1,129 @@
...
+  vlog "onload_import: $profile ($pf)"
+  # Source profile, with $@ representing the application and its options
+  shift
+  # shellcheck disable=SC1090
+  . "$pf"
+  vlog "onload_import-done: $profile"
+}
+
+######################################################################
+# main()
```
ivatet-amd, line 83: I wanted to suggest making it a real `main()` called at the end of the script, but then realised it's still a copied part of the script!

scripts/profiles/process_profile.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+
+me=$(basename "$0")
```
ivatet-amd, line 5: Likewise, shall we add the SPDX headers and change MIT to BSD-2-Clause?

scripts/profiles/process_profile.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
```
ivatet-amd, line 1: Would you prefer a more detailed name like profile_to_configmap.sh?

ivatet-amd (commented):
Thank you for implementing the "dynamic" approach, where we don't copy or manually pre-process the Onload profiles in this repo.

Sorry for the nit-picking:
```diff
-If you want to apply the generate ConfigMap straight away, then you can simply run:
+If you want to apply the generated ConfigMap straight away, then you can simply run:
```

Would you like to update the README.md too?

ivatet-amd (approved):

PR #88, opened 2023-11-28
ON-15438: Document sfptpd deployment using Kustomize

pcolledg-amd:
Uses pure Kustomize to enable user environment-specific customisation and enable us to provide further use case and version updates without modifying core config.

Smoke test overlay works in minikube.

In review until:
- [ ] in-cluster tests complete

Outputs:
```sh
kubernetes-onload/config/samples/sfptpd$ kubectl kustomize overlays/example-sf0/
```

```yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    node-role.kubernetes.io/ptp: ""
  name: sfptpd
---
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    node-role.kubernetes.io/ptp: ""
  name: sfptpd
  namespace: sfptpd
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    node-role.kubernetes.io/ptp: ""
  name: sfptpd
  namespace: sfptpd
rules:
- apiGroups:
  - security.openshift.io
  resourceNames:
  - privileged
  resources:
  - securitycontextconstraints
  verbs:
  - use
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  labels:
    node-role.kubernetes.io/ptp: ""
  name: sfptpd
  namespace: sfptpd
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: sfptpd
subjects:
- kind: ServiceAccount
  name: sfptpd
  namespace: sfptpd
---
apiVersion: v1
data:
  sfptpd.cfg: |
    [general]
    sync_module ptp ptp1

    # Log to stdout/stderr. The recommended way to deploy would be only logging
    # the message log, to stderr, and placing an adapter program in the pod to
    # process RT JSON stats and make them available to a time series database.
    message_log stderr
    stats_log stdout

    [ptp1]
    ptp_domain 0
    interface sf0
kind: ConfigMap
metadata:
  labels:
    node-role.kubernetes.io/ptp: ""
  name: sfptpd-config-c65fbfbchb
  namespace: sfptpd
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    node-role.kubernetes.io/ptp: ""
  name: sfptpd
  namespace: sfptpd
spec:
  selector:
    matchLabels:
      name: sfptpd
  template:
    metadata:
      labels:
        name: sfptpd
    spec:
      containers:
      - args:
        - -f
        - /etc/sfptpd/sfptpd.cfg
        image: docker.io/onload/sfptpd:3.7.1.1000
        imagePullPolicy: Always
        name: sfptpd
        securityContext:
          privileged: true
        volumeMounts:
        - mountPath: /etc/sfptpd
          name: etc-sfptpd
      hostNetwork: true
      nodeSelector:
        node-role.kubernetes.io/ptp: ""
        node-role.kubernetes.io/worker: ""
      serviceAccountName: sfptpd
      volumes:
      - configMap:
          name: sfptpd-config-c65fbfbchb
        name: etc-sfptpd
```

### DIFF
--- a/config/samples/sfptpd/README.md
+++ b/config/samples/sfptpd/README.md
@@ -1,0 +1,61 @@
+
+# Solarflare Enhanced PTP Daemon (sfptpd)
+
+To run `sfptpd` in a Kubernetes cluster, deploy the `sfptpd` privileged
+container to all nodes receiving PTP signal and
+[disable the node's existing time services](#disable-default-time-source).
+
+This directory provides a Kustomize formatted base
+([DaemonSet](base/daemonset.yaml)) and example overlays.
+
+(The Onload Operator does not presently manage `sfptpd`; please refer to
+Release Notes for details on expressing interest in this feature.)
+
+## Quickstart
+
+The example Kustomize overlay
+[overlays/example-bond0/kustomization.yaml](overlays/example-bond0/kustomization.yaml)
+would deploy the following to a namespace called `sfptpd`:
+
+1. A DaemonSet deploying the `sfptpd` container to worker nodes labelled
+   `node-role.kubernetes.io/ptp`.
+2. A [sfptpd.cfg](overlays/example-bond0/sfptpd.cfg) config file specifying
+   the pod's PTP ethernet interface name as `bond0`. (A bond is not required.)
+
+To use in your own cluster, make a copy and apply:
+
+```sh
+cp -r overlays/example-bond0 overlays/my-cluster
+$EDITOR overlays/my-cluster/kustomization.yaml
+$EDITOR overlays/my-cluster/sfptpd.cfg
+kubectl label nodes <your-ptp-node> node-role.kubernetes.io/ptp=
+kubectl apply -k overlays/my-cluster
+```
+
+Ensure you have also disabled the node's existing time services.
+
+## Disable default time source
+
+A node must have one time source only. sfptpd should replace services such
+as `ntpd` and `chronyd` if they manage the local host's time.
+
+If you are running OpenShift, follow the [Redhat documentation on disabling chronyd.service](https://docs.openshift.com/container-platform/4.10/post_installation_configuration/machine-configuration-tasks.html#cnf-disable-chronyd_post-install-machine-configuration-tasks).
+Ensure the resulting MachineConfigPool exactly selects the nodes specified
+above; if using the Quickstart's node label of `node-role.kubernetes.io/ptp`,
+use a NodeSelector matching this label.
+
+## Source Code & Documentation
+
+Container images with compiled binaries are provided. To build the container
+afresh, refer to
+[sfptpd on GitHub](https://github.com/Xilinx-CNS/sfptpd/#building-a-container-image).
+
+Example config (.cfg) files are available from
+[sfptpd on GitHub](https://github.com/Xilinx-CNS/sfptpd/tree/master/config).
+
+Full documentation is available in the
+[Enhanced PTP User Guide](https://docs.xilinx.com/r/en-US/ug1602-ptp-user)
+
+## Footnotes
+
+(c) Copyright 2023 Advanced Micro Devices, Inc.
--- a/config/samples/sfptpd/base/daemonset.yaml
+++ b/config/samples/sfptpd/base/daemonset.yaml
@@ -1,0 +1,64 @@
+# SPDX-License-Identifier: MIT
+# (c) Copyright 2023 Advanced Micro Devices, Inc.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sfptpd
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: sfptpd
+rules:
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
+  resourceNames:
+  - privileged
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: sfptpd
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: sfptpd
+subjects:
+- kind: ServiceAccount
+  name: sfptpd
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: sfptpd
+spec:
+  selector:
+    matchLabels:
+      name: sfptpd
+  template:
+    metadata:
+      labels:
+        name: sfptpd
+    spec:
+      hostNetwork: true
+      serviceAccountName: sfptpd
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+      containers:
+      - image: onload/sfptpd:latest
+        name: sfptpd
+        imagePullPolicy: Always
+        volumeMounts:
+        - name: etc-sfptpd
+          mountPath: "/etc/sfptpd"
+        securityContext:
+          privileged: true
+        args: ["-f", "/etc/sfptpd/sfptpd.cfg"]
+      volumes:
+      - name: etc-sfptpd
+        configMap:
+          name: sfptpd-config
--- a/config/samples/sfptpd/base/kustomization.yaml
+++ b/config/samples/sfptpd/base/kustomization.yaml
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- daemonset.yaml
--- a/config/samples/sfptpd/overlays/example-bond0/kustomization.yaml
+++ b/config/samples/sfptpd/overlays/example-bond0/kustomization.yaml
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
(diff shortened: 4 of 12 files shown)


config/samples/sfptpd/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+To run `sfptpd` in a Kubernetes cluster, deploy the `sfptpd` privileged container to the node receiving PTP signal.
+
+This directory provides Kustomize formatted:
+
+* base ([DaemonSet](base/daemonset.yaml)),
+* components (eg. [MachineConfig for chronyd](chronyd/machineconfig-4.8.0/99-worker-chronyd.yaml), [ConfigMap *sfptpd.conf*](conf/ptp1/configmap.yaml)), and
+* example overlay.
+
+(The Onload Operator does not presently manage `sfptpd`; please inform your AMD account manager know if this would be a valued feature addition.)
```
abower-amd, line 12: `s,know,now,g`

config/samples/sfptpd/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Quickstart
+
+The example Kustomize overlay [default-ocp/kustomization.yaml](default-ocp/kustomization.yaml) is tailored to OpenShift's MachineConfig and will template the following:
+
+1. A DaemonSet deploying the `sfptpd` container, specifying the pod's PTP interface name as `sf0` and using a *sfptpd.conf* file specifying `ptp1` kernel device.
```
abower-amd, line 18: I think it would be better to keep the interface name spec to the config file which needs supplying anyway and not vary the command line, where the specified interface is in any case only a default for where it isn't specified in the config file.

config/samples/sfptpd/base/daemonset.yaml
```diff
@@ -0,0 +1,66 @@
...
+      containers:
+      - image: onload/sfptpd:latest
+        name: sfptpd
+        imagePullPolicy: Always
+        volumeMounts:
+        - name: etc-sfptpd
+          mountPath: "/etc/sfptpd"
+        securityContext:
+          privileged: true
+        args: ["--no-daemon", "-f", "/etc/sfptpd/ptp1.conf"]
```
abower-amd, line 62: `--no-daemon` is redundant because it is already specified in the environment file but I think it might be a good idea to keep it here too.

config/samples/sfptpd/conf/ptp1/configmap.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+    # Log to stdout/stderr. The recommended way to deploy would be only logging
+    # the message log, to stderr, and placing an adapter program in the pod to
+    # process RT JSON stats and make them available to a time series database.
+    message_log stderr
+    stats_log stdout
+
+    [ptp1]
+    ptp_domain 0
+    transport ipv6
```
abower-amd, line 22: As this will certainly need customising by the user I suggest adding an example interface stanza and removing the `ipv6` example which was only there previously to provide out-of-the-box action using automatic link local address but is very unlikely to match the user's requirements.

config/samples/sfptpd/default-ocp/assert.sh
```diff
@@ -0,0 +1,21 @@
...
+    exit 1
+  else
+    echo "$doc_type/$doc_name: $key == $value"
+  fi
+}
+
+
+assert 'MachineConfig' '99-worker-chrony-disable' 'metadata.labels.timesource' 'sfptpd'
+assert 'DaemonSet' 'sfptpd' 'spec.template.spec.nodeSelector.timesource' 'sfptpd'
+assert 'DaemonSet' 'sfptpd' 'spec.template.spec.containers[0].args[-1]' 'sf0'
```
abower-amd, line 21: We have removed this argument.

config/samples/sfptpd/default-ocp/kustomization.yaml
```diff
@@ -0,0 +1,42 @@
...
+- ../chronyd/machineconfig-4.8.0
+
+# Corresponds to namespace.yaml
+namespace: sfptpd
+
+# Latest stable
+images:
+- name: onload/sfptpd
+  newName: docker.io/onload/sfptpd
+  newTag: 3.8.0.1000
```
abower-amd, line 21: I suggest we point to a release candidate on the stable `v3_7` branch after backporting the workflows as `3.8` is unstable.

config/samples/sfptpd/default-ocp/kustomization.yaml
```diff
@@ -0,0 +1,42 @@
...
+patches:
+- patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/args/-1
+      value: -i
+    - op: add
+      path: /spec/template/spec/containers/0/args/-1
+      value: sf0
+  target:
+    kind: DaemonSet
```
abower-amd, line 42: Interface should be configured in the config file not the command line.

config/samples/sfptpd/README.md
abower-amd, line 18: Where's this `ptp1`? The `ptp1` specified in the config file is scoped only to the config file, it is the name of a 'sync instance'. The `/dev/ptp*` kernel devices I believe are already mapped into the pod?

abower-amd (changes requested):

krishd-amd:
We should probably settle on how we are going to approach yaml generation, templating and butane.

tcrawley-xilinx (commented):
Do we really need to butane files that differ only with the ignition version? In the latest version of #86 there is a `compile_butane.sh` script that looks like it could be applicable here as well.

What is the expected usage of `./assert.sh`? Are user expected to use it or is it for internal/CI usage?

abower-amd (approved):
Like the three examples! Good way of illustrating what the user can vary!

config/samples/sfptpd/README.md
```diff
@@ -0,0 +1,46 @@
...
+This directory provides a Kustomize formatted base ([DaemonSet](base/daemonset.yaml)) and example overlays.
+
+(The Onload Operator does not presently manage `sfptpd`; please inform your AMD Account Manager if this would be a valued feature addition.)
+
+## Quickstart
+
+The example Kustomize overlay [overlays/example-sf0/kustomization.yaml](overlays/example-sf0/kustomization.yaml) would deploy the following to a namespace called `sfptpd`:
+
+1. A DaemonSet deploying the `sfptpd` container to worker nodes labelled `node-role.kubernetes.io/ptp`.
+2. A [sfptpd.cfg](overlays/example-sf0/sfptpd.cfg) config file specifying the pod's PTP ethernet interface name as `sf0`.
```
krishd-amd, line 15: It would be useful to have an instruction here to perhaps clarify how to find sf0? I think on the pods created, the accelerated onload interface is net1?

config/samples/sfptpd/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+To run `sfptpd` in a Kubernetes cluster, deploy the `sfptpd` privileged container to all nodes receiving PTP signal and [disable the node's existing time services](#disable-default-time-source).
+
+This directory provides a Kustomize formatted base ([DaemonSet](base/daemonset.yaml)) and example overlays.
+
+(The Onload Operator does not presently manage `sfptpd`; please inform your AMD Account Manager if this would be a valued feature addition.)
+
+## Quickstart
+
+The example Kustomize overlay [overlays/example-sf0/kustomization.yaml](overlays/example-sf0/kustomization.yaml) would deploy the following to a namespace called `sfptpd`:
```
krishd-amd, line 12: This section I think could do with some clarification.

config/samples/sfptpd/README.md
krishd-amd: Is this meant to be borrowed across configs, from the example posted in slack I think this is meant to refer to the bond created on the host.

If so we have a prerequisite probably that all the hosts have the same interface name here I think? That should be clarified in the documentation.

config/samples/sfptpd/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+1. A DaemonSet deploying the `sfptpd` container to worker nodes labelled `node-role.kubernetes.io/ptp`.
+2. A [sfptpd.cfg](overlays/example-sf0/sfptpd.cfg) config file specifying the pod's PTP ethernet interface name as `sf0`.
+
+To use in your own cluster, make a copy and apply:
+
```
krishd-amd, line 18: Can there be a step here to just check if things are working correctly?

krishd-amd (changes requested):

config/samples/sfptpd/overlays/example-sf0/sfptpd.cfg
```diff
@@ -0,0 +1,15 @@
...
+
+# Log to stdout/stderr. The recommended way to deploy would be only logging
+# the message log, to stderr, and placing an adapter program in the pod to
+# process RT JSON stats and make them available to a time series database.
+message_log stderr
+stats_log stdout
+
+[ptp1]
+ptp_domain 0
+interface sf0
```
krishd-amd, line 15: This I think needs a ptp_mode master line.

abower-amd: No it doesn't `example-sf0` is for customers to use/tailor for their environment which will have an external third party grandmaster. Your change converts this into a working alternative to the "link-local" example which therefore needs changing and renaming.

config/samples/sfptpd/README.md
abower-amd: You could have an `exec` command example that did `cat topology` or something.

ivatet-amd (approved):
Looks good! +1 for @krishd-amd's `cp -r` suggestion.

krishd-amd:
I have put my changes here if needed https://github.com/krishd-amd/kubernetes-onload/tree/sfptpd-doc

PR #100, opened 2023-12-08
Set Device Plugin Image at controller level

tcrawley-xilinx:
Previously the device plugin image would be set per Onload CR, this could potentially lead to confusion since (we expect) most users won't need to change the device plugin image nor have fine-grained control of the image. Additionally, since we have built the controller with certain assumptions of the device plugin image it should lower the chances that a user sets this to an incompatible value.

I have left a field for the ImagePullPolicy in the `DevicePluginSpec` struct, though I'm wondering how necessary that is.

## Testing done
Tested in virtualised cluster, when an onload CR is created it creates the device plugin correctly.

Since the value of the device plugin image is set at operator deploy time, I have found that the best way to deploy the operator is with the command
```
make deploy IMG=... DEVICE_IMG=...
```

### DIFF
--- a/controllers/onload_controller.go
+++ b/controllers/onload_controller.go
@@ -37,7 +37,8 @@ import (
 // OnloadReconciler reconciles a Onload object
 type OnloadReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme            *runtime.Scheme
+	DevicePluginImage string
 }
 
 //+kubebuilder:rbac:groups=onload.amd.com,resources=onloads,verbs=get;list;watch;create;update;patch;delete
@@ -910,7 +911,7 @@ func (r *OnloadReconciler) createDevicePluginDaemonSet(
 
 	devicePluginContainer := corev1.Container{
 		Name:            "device-plugin",
-		Image:           onload.Spec.DevicePlugin.DevicePluginImage,
+		Image:           r.DevicePluginImage,
 		ImagePullPolicy: onload.Spec.DevicePlugin.ImagePullPolicy,
 		SecurityContext: &corev1.SecurityContext{
 			Privileged: ptr.To(true),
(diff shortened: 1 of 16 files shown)


controllers/onload_controller.go
```diff
@@ -858,12 +859,19 @@ func createModule(
...
+const devicePluginImageEnv = "DEVICE_PLUGIN_IMG"
 
 func (r *OnloadReconciler) createDevicePluginDaemonSet(
 	ctx context.Context, onload *onloadv1alpha1.Onload,
 ) (*ctrl.Result, error) {
 	log := log.FromContext(ctx)
 
+	devicePluginImage := os.Getenv(devicePluginImageEnv)
+	if devicePluginImage == "" {
+		return nil, fmt.Errorf("%q environment variable empty or not set."+
```
ivatet-amd, line 871: Maybe even refuse to start the controller if `DEVICE_PLUGIN_IMG` is not set?

One example is KMM in `main()`:
```go
workerImage := cmd.GetEnvOrFatalError("RELATED_IMAGES_WORKER", setupLogger)
```
(https://github.com/kubernetes-sigs/kernel-module-management/blob/940e25e11ed701e6c218577066e29cadb4d9a62d/cmd/manager/main.go#L87)

ivatet-amd (approved):
At first, I wouldn't say I liked the idea of removing the device plugin image field from CRD because, from time to time, I edit it in the Onload CR and reapply without redeploying the controller. However, the ENV approach still looks good to me, and, for instance, KMM seems to do the same thing!

Also, it seems reasonable **not** to expose it to the user because we don't want to maintain compatibility between the controller and the device plugin. Only 1-1 mapping!

On second thought, it looks good to me, but please take a look at the inline suggestion.

ivatet-amd:
https://github.com/Xilinx-CNS/kubernetes-onload/blob/master/DEVELOPING.md refers to `devicePluginImage`: also needs updating.

controllers/onload_controller.go
tcrawley-xilinx: Done as a new commit, if you're happy with it I'll squash it before merging the PR

main.go
```diff
@@ -58,6 +59,17 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	// Probably overkill given that a check exists the the controllers package,
```
ivatet-amd, line 62:
```suggestion
	// Probably overkill given that a check exists the controllers package,
```
Actually, shall we simply remove the entire comment because (1) there is no check in the controller package, (2) it's certainly not overkill and (3) it seems the most reasonable (the earliest) place to check the env?

ivatet-amd (approved):

PR #101, opened 2023-12-08
cleanup: Add ContainerFileNotices for Ubuntu & MachineConfig

ivatet-amd:
There are two new Dockefiles: one to generate SFC MachineConfig, and another is an Onload kernel module builder for the Ubuntu kernels. This patch adds the corresponding ContainerFileNotice even though the latter image is a template.

---

Re on Dockerfiles:

- I chose the specific version of Butane (v0.19.0), which must be the latest `release` tag.
- Ubuntu is borderline, as we don't build these images, but added for completeness.


tcrawley-xilinx (approved):

ivatet-amd:
@pcolledg-amd Thanks, I've expanded the commit message. 👍

ivatet-amd:
I've listed them in the body of the commit message, but your version works for me too.

PR #108, opened 2023-12-14
ON-15462: Docs - CRD from `kubectl explain`

pcolledg-amd:
`onload.spec.devicePlugin.devicePluginImage` omitted in anticipation of #100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ a [Onload Custom Resource (CR)](#onload-custom-resource-cr).
 
 Instruct the Onload Operator to deploy the components necessary for accelerating workload pods by deploying
 a `Onload` *kind* of Custom Resource (CR).
+For its schema, refer to [`Onload` Custom Resource Definition (CRD)](docs/crd.md).
 
 If your cluster is internet-connected OpenShift and you want to use in-cluster builds with the current version
 of OpenOnload, run:
--- a/api/v1alpha1/onload_types.go
+++ b/api/v1alpha1/onload_types.go
@@ -9,6 +9,7 @@ import (
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// NOTE: Update corresponding docs/crd.md
 
 // Currently unimplemented
 type SFCSpec struct {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -147,6 +147,7 @@ a [Onload Custom Resource (CR)](#onload-custom-resource-cr).
 
 Instruct the Onload Operator to deploy the components necessary for accelerating workload pods by deploying
 a `Onload` *kind* of Custom Resource (CR).
+For its schema, refer to [`Onload` Custom Resource Definition (CRD)](docs/crd.md).
```
ivatet-amd, line 150: I see why the CRD doc is helpful, but it creates a maintenance burden and is liable to become forgotten and stale! I think it is good for a more static form of documentation, like the user guide shipped separately, where we can put a snapshot of the CRD.

How about offering the `kubectl explain` flavours to the user, e.g. `kubectl explain onload` and `kubectl explain onload.spec`, so that they check the CRD themselves?

ivatet-amd: Please do not get upset with my comments. It is purely an opinion. Moreover, I suggest others share theirs so we land on the same page after a friendly discussion. Let me try to clarify it.

I wanted to say that it seems wrong (to me) to copy and paste the utility output into the doc as-is. It is not friendly, not formatted as Markdown and limits the audience. The users who are willing to look at the plain `kubectl explain` output, are also skilled in reading the Golang definitions because this documentation (often based on my observations) becomes outdated. Hence, a plain `kubectl explain` adds little to no value to these people (in my opinion, again).

I agree that users want to understand the features upfront, so the suggestion `kubectl explain` does not help progress. Instead, let me suggest either:

- (I believe you are referring to it in the Jira comments:) Autogenerated CRD documentation in the repo, for instance, with one `make` invocation. https://github.com/openshift/cluster-node-tuning-operator/blob/master/docs/performanceprofile/performance_profile.md is auto-generated. This way, we can have automation check that the CRD docs are up-to-date.
- A friendly, human-written page with examples. For instance, https://github.com/kubernetes-sigs/kernel-module-management/blob/main/docs/mkdocs/documentation/deploy_kmod.md. I see it is human-written, and it makes me instantly assume it can be outdated, but I see plenty of value in it anyway thanks to the extensive comments with the implementation overview.

As for the user guide, many eyeballs look at it before we ship it, and the outdated parts are more likely to get noticed.

PR #114, opened 2023-12-18
cleanup: v3.0.0 `make bundle` with devicePluginImage updates

pcolledg-amd:
Proxy image patch [as per KMM](https://github.com/rh-ecosystem-edge/kernel-module-management/blob/cd3cd0439c7a045a90f7d6c54f046464781dc189/config/manager-base/kustomization.yaml#L15).

Doc fixes missed in #100.

CSV links & versions updated for release, mostly cosmetic.

Dry run tested only.


tcrawley-xilinx (approved):
Changes look plausible. How did you update the bundle stuff? Did you update it manually or regenerate the file?

bundle/metadata/dependencies.yaml
```diff
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: MIT
 # SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
```
tcrawley-xilinx, line 4: Double copyright headers. I don't think this file is auto-generated, so it is worth checking it doesn't re-add the headers each time the bundle is generated.

config/samples/kustomization.yaml
```diff
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: MIT
 # SPDX-FileCopyrightText: (c) Copyright 2023 Advanced Micro Devices, Inc.
+
+# Bundled into CSV's `alm-examples`
```
tcrawley-xilinx, line 4: Could you elaborate on this? Or, if possible could it be possible to link it to a filepath like `config/samples/onload/kustomization.yaml` rather than here? I'm not completely comfortable with having the root of all samples only point to onload when when there are other types of samples included (sfptpd, sfnettest, etc.)

config/samples/kustomization.yaml
```diff
@@ -1,4 +1,6 @@
...
 resources:
-- base
+- onload/overlays/in-cluster-build-ocp
```
ivatet-amd, line 6: Do you reckon it's a good timing to enable SFC (uncomment `sfc: {}`) in the overlay file?

bundle/manifests/onload-operator.clusterserviceversion.yaml
```diff
@@ -10,42 +10,50 @@ metadata:
...
+                        "name": "ONLOAD_BUILD_PARAMS",
+                        "value": ""
+                      },
+                      {
+                        "name": "ONLOAD_SOURCE",
+                        "value": "docker.io/onload/onload-source:8.1.2.26"
+                      }
+                    ],
+                    "dockerfileConfigMap": {
+                      "name": "onload-module-dockerfile-ch29bghg54"
```
ivatet-amd, line 35: Why do we want `ch29bghg54` in the resource and file name?

ivatet-amd (approved):

config/samples/kustomization.yaml
tcrawley-xilinx: If its an external constraint then i don't have a real issue with it.

tcrawley-xilinx (approved):

PR #119, opened 2024-01-08
ON-15256: Report topology info from device plugin

tcrawley-xilinx:
### [1/2] ON-15180: Read sysfs to check for NICs

Rather than rely on an external tool (lshw) which requires extra
dependencies in container images and doesn't provide an simple way of
determining the vendor of a nic, it would be better to search through
sysfs directly (since that is basically all lshw is doing). This
approach also allows us to easily get other information from sysfs (such
as numa nodes).

### [2/2] ON-15256: Report topology info from device plugin

The device plugin will now try to read the numa information from sysfs
while finding the nics present in the system. The device plugin will
then add all of the present numa node information to the list of
advertised devices. Because the device plugin is only advertising a
pseudo-device it is unclear how to proceed when multiple nics are
present on different nodes.

## Testing done

Some manual testing in the cluster, no unit tests (yet) trying to think of a way of mocking sysfs files.


pkg/deviceplugin/nic.go
```diff
@@ -3,69 +3,97 @@
...
+		return false
+	}
+
+	vendor := strings.TrimSuffix(string(data), "\n")
+	return vendor == solarflareVendor
+}
 
-	var interfaces []string
+// Get numa node from sysfs files.
+// -1 means no specific numa node / unknown
```
ivatet-amd, line 48: -1 is also conveniently aligned with the kernel's `NUMA_NO_NODE` macro! 👍

pkg/deviceplugin/nic.go
```diff
@@ -3,69 +3,97 @@
...
+	} else if err != nil {
+		glog.Errorf("Error reading %s (%v)",
+			path.Join(devicePath, "device", "vendor"), err)
+		return -1
+	}
+	numaString := strings.TrimSuffix(string(data), "\n")
+	node, err := strconv.ParseInt(numaString, 10, 64)
+	if err != nil {
+		glog.Errorf("Error parse int from string %s (%v)",
+			string(data), err)
```
ivatet-amd, line 63: nit:

```suggestion
			numaString, err)
```

`string(data)` is also OK, but that's not what we give to `strconv.ParseInt()`.

pkg/deviceplugin/manager.go
```diff
@@ -92,13 +101,28 @@ func NewNicManager(
 
 // Initialises the set of devices to advertise to kubernetes
 func (manager *NicManager) initDevices() {
+
+	// Gets a list of all numa nodes of which there is an associated sfc nic,
+	// this isn't particularly helpful when you have nics on different numa
+	// nodes which are intended for different purposes, but this is basically a
+	// pathological case due to how we only advertise an "onload" device rather
+	// than a "real" one.
+	numaNodes := []*pluginapi.NUMANode{}
```
ivatet-amd, line 110: Do we need to ensure uniqueness in the `numaNodes` slice? Either way, it would be nice to capture it as a unit test...

pkg/deviceplugin/manager.go
```diff
@@ -40,12 +40,17 @@ var DefaultConfig = NicManagerConfig{
 	NeedNic:        true,
 }
 
+type nic struct {
+	name string
+	numa int64
```
ivatet-amd, line 45: nit: Maybe `int16` or `int`? The maximum `NODES_SHIFT` value on x86 seems to be 10, i.e. 1K NUMA nodes.

tcrawley-xilinx: Initially I had it as an `int`, but the spec for a [`NUMANode`](https://github.com/kubernetes/kubelet/blob/v0.28.3/pkg/apis/deviceplugin/v1beta1/api.pb.go#L308) from the deviceplugin api uses `int64`, so I thought being consistent was probably a easier than dealing with the conversions. I'm not anticipating running this on a machine with over 4 billion numa nodes!

pkg/deviceplugin/nic.go
```diff
@@ -3,69 +3,97 @@
...
-	// /0/100/1b/0         enp2s0f0   network        XtremeScale SFC9250 10/25/40/50/100G Ethernet Controller
-	// /0/100/1b/0.1       enp2s0f1   network        XtremeScale SFC9250 10/25/40/50/100G Ethernet Controller
-	// /0/100/1c.1/0       eno1       network        NetXtreme BCM5720 Gigabit Ethernet PCIe
-	// /0/100/1c.1/0.1     eno2       network        NetXtreme BCM5720 Gigabit Ethernet PCIe
+const (
+	sysClassNetPath  = "/sys/class/net/"
+	solarflareVendor = "0x1924"
+)
+
+func isSFCNic(devicePath string) bool {
```
krishd-amd, line 20: Would it be worth using isOnloadableNic or isSupportedNic?

Just in case we are asked to support another vendor sometime in the future.

tcrawley-xilinx:
> Consider using [ghw go library](https://github.com/jaypipes/ghw#network) to avoid the mass of filesystem handling boilerplate -- it supports PCIe vendors, network devices, and topology.

I had a look into it, and it would work if we just cared about netdevices, but I don't think it will work for our purposes due to how it deals with pci devices. Rather than reading the files as I have done in this change, I wants to use a pci database (either from the file system or through the network).
When testing this is in the cluster it would fail to enumerate the pci devices, complaining about the lack of the pci database.

ivatet-amd (approved):
Looks good, thank you!

I would also suggest filing a Jira ticket for the `queryNics()` unit test coverage. For instance, we could create a mock sysfs tree in the repo and feed it to the tests by replacing `sysClassNetPath` (or similarly, whatever suits better).

tcrawley-xilinx:
> I would also suggest filing a Jira ticket for the `queryNics()` unit test coverage.

ON-15228?

> For instance, we could create a mock sysfs tree in the repo and feed it to the tests by replacing `sysClassNetPath` (or similarly, whatever suits better).

I do want to increase test coverage for parts of the code that interact with the filesystem (this, `mounts.go` and maybe some bits in `onload-worker`). Another idea I had was to create a wrapper interface around common filesystem calls (read, write, stat etc) then provide a mocked implementation so that tests don't have to interact with the real filesystem.

ivatet-amd:
Awesome! [ON-15228](https://jira.xilinx.com/browse/ON-15228) seems like the perfect fit. I bet you're already typing a comment over there with suggestions to improve the test coverage if anyone else wants to take a look at it unless you're planning to implement it in this PR (I assumed not, given it's ready for review)?!

deviceplugin.Dockerfile
```diff
@@ -22,7 +22,6 @@ COPY LICENSE /app/LICENSE
 RUN CGO_ENABLED=0 make device-plugin-build worker-build
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.9
-RUN microdnf install -y lshw-B.02.19.2 && microdnf clean all
```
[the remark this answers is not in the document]
tcrawley-xilinx: Yes it does! I did some testing initially using `scratch` as the base image, but that would fail with the initContainer/lifecycle commands. I had forgotten about `ubi-micro` 🤦‍♂️

PR #120, opened 2024-01-11
ON-15512: Don't fail with non cri-o container runtime

tcrawley-xilinx:
Any container runtime that is compatible with the cri should work fine since we are using `crictl` which is designed to run against the standard.

## Testing done
Ran in-cluster, as expected no warning message.


ivatet-amd (approved):
It's nice that we can reuse `crictl` for containerd which comes with its own `ctr`.

PR #121, opened 2024-01-11
ON-15464: Recommended Labels and onload-version rename

pcolledg-amd:
For upcoming troubleshooting guide.

* Namespaced the `onload-version` label. Not required here but now sticks out when compared alongside other identifiers. Clarifies that it's not for general users to touch arbitrarily. Candidate for adding to upgrade docs.
* Added `/part-of` and `/managed-by` [Recommended Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for much easier selection in troubleshooting guide, eliminating hacky/greedy grepping or awkward 'field selectors'. Also added `/component` and `/instance` for clarity.
* Renamed `onload-onload-device-plugin-init` to just `init` as name need only be unique within pod
* Removed redundant test target from `make device-plugin-docker-push`

The `onload.amd.com/name` label is unchanged, only a DRY tweak.

Would have liked to add `/version` too but there are many different product versions and some would require extra user changes, risking staleness.

### DIFF
--- a/controllers/onload_controller.go
+++ b/controllers/onload_controller.go
@@ -159,6 +159,15 @@ const kmmModuleLabelPrefix = "kmm.node.kubernetes.io/version-module"
 const onloadModuleNameSuffix = "-module"
 const sfcModuleNameSuffix = "-sfcmod"
 
+func baseLabels(name, namespace string, component string) map[string]string {
+	return map[string]string{
+		"app.kubernetes.io/managed-by": "onload-operator",
+		"app.kubernetes.io/part-of":    "onload",
+		"app.kubernetes.io/instance":   "onload-" + namespace + "-" + name,
+		"app.kubernetes.io/component":  component,
+	}
+}
+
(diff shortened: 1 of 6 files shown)


controllers/onload_controller.go
```diff
@@ -846,6 +855,7 @@ func createModule(
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      moduleName,
 			Namespace: onload.Namespace,
+			Labels:    baseLabels(onload.Name, onload.Namespace, "module"),
```
tcrawley-xilinx, line 858: Do we want the `sfc` and `onload` modules to have different labels?

tcrawley-xilinx (commented):
Just to check, are these labels consumed by anything? Or just users?
(I think the answer is just users, but want to check)

tcrawley-xilinx (approved):
A unit test would be nice, but as you acknowledge it would also be low impact

PR #127, opened 2024-01-26
ON-15555: sfptpd:stable in overlay sample

pcolledg-amd:


abower-amd:
yes, but I think there are multiple places where this needs fixing.

abower-amd (approved):

